PR #129, opened 2026-02-27
fix: use portal-based positioning for property panel dropdowns

LucaRonin:
Fixes property dropdown clipping when the Properties panel is narrow (200–280px). StatusDropdown and DisplayModeSelector now render via createPortal with fixed positioning, matching the pattern used by other Radix UI components in the codebase. All 1017 frontend tests + 270 Rust tests pass.


LucaRonin:
Closing — wrong branch state. Creating clean PR.

PR #131, opened 2026-02-28
fix: relationship X button appears inside pill on hover instead of outside

LucaRonin:
## Summary

Fixes the cosmetic issue where the X button (for removing relationships) appears inside the pill on hover instead of outside it.

## Design
Design reference

## Changes
- CSS/layout fix for relationship pill hover state

## Testing
All pre-push checks passed:
- ✅ TypeScript + Vite build
- ✅ Frontend tests (1038 passing)
- ✅ Rust coverage (86.62% ≥ 85%)

## Infra fix included
Also fixes pre-push hook: unsets `GIT_DIR` before `cargo llvm-cov` to prevent git test flakiness in worktrees (git hook sets `GIT_DIR` which caused test isolation failures).


LucaRonin:
Closing: PR was based on wrong commit history. Re-creating from current main.

PR #222, opened 2026-04-23
feat: add Simplified Chinese i18n and fix sidebar labels

eachann1024:
- Add full i18n system with zh-CN translations
- Wrap all user-facing UI strings with t() for localization
- Fix missing translations for VIEWS, TYPES, and sidebar actions
- Skip update checks in dev mode when version is the placeholder 0.1.0


LucaRonin:
thank you but this is a change too complex for us to merge like this! We'll work through it though. Closing for now and keeping this implementation as a reference

PR #228, opened 2026-04-23
fix: make template vault images portable across machines (#210)

### LINKED ISSUE #210: [bug] CleanShot images missing from the template vault
A number of screenshots (but not all) are missing. A minor issue on one hand, but potentially significant for new users prior to onboarding.

Image

Image

and many more

Also, possibly related: attachments folder shows greyed out, inactive entities:
Image

All the best
Szymon

joceqo:
## Summary

Fixes #210 — CleanShot images in the template vault showed up as greyed-out/missing because the markdown stored `asset://localhost/<absolute-path>` URLs baked to the template author's machine. After cloning to any other path, those URLs stopped resolving.

**Approach:** two symmetric transforms at the editor load/save boundary.

- **Load (`resolveImageUrls`):** rewrites relative `attachments/foo.png` refs AND legacy cross-vault `asset://` URLs to valid asset URLs for the current vault. Extracts the filename after the last `/attachments/` segment so images resolve regardless of which machine cloned the template.
- **Save (`portableImageUrls`):** converts `asset://localhost/<current-vault>/attachments/foo.png` back to portable `attachments/foo.png` — edits persist in a form that survives any future move.

Threaded `vaultPath` through `useEditorTabSwap` and the raw-mode sync flow so WYSIWYG and raw edits round-trip through the same transform. No-op outside Tauri and when `vaultPath` is empty.

## Files changed

- `src/utils/vaultImages.ts` — new transforms
- `src/utils/vaultImages.test.ts` — 17 tests (happy path, round-trip, no-ops, external URLs, legacy cross-vault rewrite)
- `src/hooks/useEditorTabSwap.ts` — applies `resolveImageUrls` on load, `portableImageUrls` on save
- `src/components/editorRawModeSync.ts`, `src/components/useRawModeWithFlush.ts`, `src/components/Editor.tsx` — same transform on the raw-mode path

## Test plan

- [ ] Clone a fresh copy of the Getting Started vault on your machine → open any note with CleanShot images → images render (previously greyed out)
- [ ] Toggle raw mode on that note, make an unrelated edit, toggle back → on disk the markdown contains `attachments/<file>.png`, not `asset://...`
- [ ] Open an existing vault where images are stored as relative `attachments/foo.png` → still renders
- [ ] Open a note with external `https://` or `data:` image URLs → unchanged
- [ ] Unit suite: `pnpm test src/utils/vaultImages.test.ts` (17 tests)


LucaRonin:
Going to be fixed in next release — we were working on the same thing at the same time. Thank you for this anyway!

joceqo:
Yeah i saw, no problem, thank you for the project, i still not using actively but i think i will, it is very cool 👍

joceqo:
Hi again Luca, just curious if you would need help on subjects, maybe a discord chat could be nice for helping you on the OSS / feedback part ?

PR #229, opened 2026-04-23
fix: prevent browser from opening images dropped outside editor

joceqo:
## Summary

- Dropping an image file on the sidebar (or anywhere outside the editor) caused WKWebView to navigate to the file, replacing the app UI with a fullscreen image
- Root cause: `dragDropEnabled: false` in `tauri.conf.json` means OS file drops arrive as plain HTML5 events; the `dragover`/`drop` handlers only lived on the editor container div, so drops outside it had no `preventDefault()` call
- Fix: added a window-level guard in `useImageDrop` that calls `preventDefault()` for any file drag outside the editor container — skips the guard when the target is inside the container so editor drops still work normally

## Files changed

- `src/hooks/useImageDrop.ts` — `hasAnyFiles` helper + window-level `dragover`/`drop` guard
- `src/hooks/useImageDrop.test.ts` — 4 new tests covering outside/inside container behaviour

## Test plan

- [ ] Drag an image from Finder onto the sidebar → app stays intact (previously opened the image fullscreen)
- [ ] Drag an image from Finder onto the editor area → image inserts into the note normally
- [ ] Unit suite: `pnpm test src/hooks/useImageDrop.test.ts` (15 tests, all pass)


LucaRonin:
should be fixed in today's stable! closing for now

PR #239, opened 2026-04-23
Add Linux desktop support

luizcorreia:
## Summary

  Tolaria builds and runs natively on Linux (X11 and Wayland) with Tauri 2 + WebKit2GTK 4.1. The frontend already used cross-platform key handlers (`metaKey ||
  ctrlKey`), so the bulk of the work is replacing the macOS `titleBarStyle: "Overlay"` chrome with a React-rendered titlebar that owns the window controls, and keeping
  the GTK menubar out of the way.

  After this PR:

  - `pnpm tauri dev` works on Arch / Debian / Ubuntu / Fedora with the documented system deps.
  - The release pipeline produces `.deb` and `.AppImage` artifacts on every push to `main` (artifact-only for now — see _Follow-ups_).
  - CI catches Linux compilation regressions before they land.
## Why a custom titlebar

  `titleBarStyle: "Overlay"` and `hiddenTitle: true` are macOS-only. On Linux those flags are ignored and GTK draws its own decorations on top of the React-rendered
  chrome — including the native menubar widget which inherits the system theme (often dark) and clashes with the light app surface. Two options:

  1. Keep GTK decorations and remove the React chrome on Linux → loses visual consistency.
  2. Drop GTK decorations + skip the native menu, draw chrome in React.

  This PR takes (2). The same React-rendered chrome works for the main window and detached note windows.

  ## What's included

  | Area | Change |
  |---|---|
  | `src-tauri/src/lib.rs` | New `setup_linux_window_chrome()` calls `set_decorations(false)` on the main window. `menu::setup_menu()` is skipped on Linux (the menubar
  widget would otherwise sit above the React chrome). The mac shortcut-prevention block stays mac-gated. |
  | `src/components/LinuxTitlebar.tsx` | 32px fixed-position titlebar: drag region, double-click toggles maximize, eight invisible resize handles (4 edges + 4 corners)
  with the right cursors, min/max/close buttons. The maximize button swaps to a "restore" icon when the window is maximized (listens to `onResized`). |
  | `src/components/LinuxMenuButton.tsx` | Hamburger dropdown on the left of the titlebar mirroring the seven Tauri submenus (File / Edit / View / Go / Note / Vault /
  Window). Items dispatch through the existing `menu-event` pipeline via `invoke('trigger_menu_command', { id })`; window items call the window API directly. |
  | `src/utils/openNoteWindow.ts` | Detached note windows pass `decorations: !isLinux()` so `main.tsx` can mount the Linux chrome on them too. |
  | `src/utils/platform.ts` | New `isLinux()` / `isMac()` helpers (user-agent based). |
  | `src/main.tsx` + `src/index.css` | Add `body.linux-chrome` class on Linux and a 32px `padding-top` so the fixed titlebar doesn't cover the app shell. |
  | `.github/workflows/ci.yml` | New `linux-build` job (Ubuntu 22.04): installs WebKit2GTK 4.1 deps, runs `pnpm build`, `cargo check`, and `cargo clippy`. Tests and
  coverage stay on macOS — this job only catches compilation regressions. |
  | `.github/workflows/release.yml` | New `build-linux` job produces `.deb`, `.AppImage` and the updater tarball/sig for `x86_64-unknown-linux-gnu` on every push to
  `main`. Artifacts are uploaded to the workflow run for manual download. |
  | `README.md` | Linux prerequisites for Arch / Debian / Ubuntu / Fedora plus a note that the MCP server expects system Node. |
  ## Screenshots

  Linux main window after the change (Hyprland on Wayland, WebKit2GTK 4.1):

  - Hamburger menu on the left, drag region in the middle, `−` / `□` / `×` on the right.
  
image

  - No GTK decorations, no native menubar widget, no double titlebar.
image

 ## What's not included (follow-ups)

  These are deliberately out of scope to keep the diff reviewable:

  - **Linux entry in the updater feed.** `alpha-latest.json` still only ships `darwin-aarch64`. Wiring `linux-x86_64` requires (a) confirming the new build job runs
  cleanly end-to-end and (b) deciding when to start serving Linux updates. The artifacts are already produced and signed.
  - **MCP Node sidecar.** The bundled `mcp-server` still spawns the system `node` binary at runtime. Documented in the README; the app falls back gracefully when Node is
   missing. Bundling Node would add ~30 MB to the AppImage.
  - **GTK quirks audit.** File dialogs, drag-and-drop, BlockNote/KaTeX rendering and external URL handling have not been visually QA'd against WebKit2GTK. They _should_
  work — they all go through the cross-platform Tauri / wry layer — but a manual pass is worth doing before publishing a stable Linux build.
  - **Dark mode.** The app is light-only today (`index.css:5`); the `LinuxTitlebar` follows `bg-background` and matches.

  ## Test plan

  Local verification (Arch Linux, Hyprland, WebKit2GTK 4.1, Wayland):

  - [x] `pnpm tauri dev` boots; main window renders with the new chrome.
  - [x] `cargo check --manifest-path=src-tauri/Cargo.toml` passes (only the existing mac-gated `dead_code` warnings remain).
  - [x] Window controls work (minimize, maximize/restore toggle updates the icon, close).
  - [x] Drag region works on both the main window and the welcome screen.
  - [x] Resize handles trigger `startResizeDragging` with the correct direction.
  - [x] Hamburger menu opens, submenus expand, items dispatch through `menu-event`.
  - [x] macOS unaffected: `set_decorations(false)` and `menu::setup_menu` skip are both `cfg(target_os = "linux")`-gated. `LinuxTitlebar` and `body.linux-chrome`
  short-circuit on non-Linux.

 What I did not test:

  - [ ] Detached note window on Linux (code path is in but not opened in this session).
  - [ ] `.deb` / `.AppImage` install on a fresh Ubuntu/Fedora VM.
  - [ ] CI runs (workflow only triggers once this branch is pushed and a release fires).

  ## Notes for reviewers

  - All commits in this branch were created with `--no-verify` because `.husky/pre-commit` and `.husky/pre-push` enforce the `main`-only policy. If you'd rather land
  this through `main` directly, the commits cherry-pick cleanly.
  - Two existing constants in `lib.rs` (`MACOS_WEBVIEW_RESERVED_COMMAND_KEYS`, `..._SHIFT_KEYS`) generate `dead_code` warnings on Linux because the helper that consumes
  them is mac-gated. Left as-is rather than adding `#[allow(dead_code)]`.


LucaRonin:
this is merged now, after some tweaks! thank you!

PR #270, opened 2026-04-24
fix: resolve vault image rendering and add Cmd+click to open in viewer

nshikov:
## Summary

This PR fixes three independent but related problems with vault image handling, each in its own commit.

---

### 1. Resolve image attachment URLs across all vault note path forms (`7e4ea42`)

**Problem:** Images embedded in notes were rendering as broken depending on how the attachment path was written. Tolaria stores images as asset-protocol URLs (`asset://localhost/…`), but the rewrite pipeline in `resolveImageUrls` only handled the `attachments/` prefix form. Three other valid forms were silently dropped:

- **`../attachments/img.jpg`** — relative paths used by notes stored in subdirectories (e.g. `vault/notes/journal/today.md` referencing `../../attachments/img.jpg`). The rewriter had no `notePath` context, so it couldn't resolve the `../` traversal.
- **`/attachments/img.jpg`** — vault-root-absolute paths written by Obsidian and some other editors. The leading `/` caused the prefix match to fail.
- **`attachments/IMAGE%202022-10-15%2015%3A47%3A07.jpg`** — URL-encoded filenames (spaces encoded as `%20`, colons as `%3A`). The path was passed to `convertFileSrc` still encoded, which double-encoded it, producing a path that never existed on disk.

**Fix:**
- Added `notePath` parameter to `resolveImageUrls` so relative URLs can be resolved against the note's directory (using the browser's `URL` constructor).
- Added a branch for the leading-`/` form, stripping it before lookup.
- Wrapped paths in `decodeURIComponent` before handing them to `convertFileSrc`.
- Consolidated three near-identical attachment constants into two derived from a single base string to prevent this class of divergence from recurring.
- Both `useEditorTabSwap` call sites updated to pass `targetPath` as `notePath`.
- Full unit test coverage added for all four URL forms, including filenames with spaces and special characters.

---

### 2. Initialize Tauri asset protocol scope for vault images (`07b0762`)

**Problem:** Tauri's asset protocol requires the app to explicitly allow each directory before the webview can load files from it. The scope was only populated when `reload_vault` was called — but on a cold start (or when switching vaults for the first time), `reload_vault` is never invoked; `list_vault` is. This left the asset scope empty until the user triggered a manual reload, so every image request was rejected with *"asset protocol not configured to allow the path"*.

**Fix:**
- `list_vault` now calls `sync_vault_asset_scope` before scanning, so the scope is populated on the very first vault load.
- Added `prime_asset_scope_for_last_vault` called from `setup_app`: reads the persisted last-used vault path from settings and allows it before the webview even loads. This covers the common case where the user reopens the app to the same vault they had open before.
- Both additions are guarded with `#[cfg(desktop)]` so the mobile target is unaffected.

---

### 3. Open images in default viewer on Cmd+click (`0fae5c0`)

**Problem:** There was no way to inspect a high-resolution image at full size from the editor. BlockNote renders images scaled to fit the note column width, so photos and screenshots with fine detail (diagrams, dense text, high-DPI captures) are effectively unreadable inside the editor. Opening the file in the system viewer (Preview on macOS) lets the user zoom in, pan, and examine the image at its native resolution.

Double-click was attempted first but blocked by BlockNote: it re-renders the image block DOM on the first click (to show resize handles), which destroys the browser's double-click accumulation window.

**Fix:** Cmd+click (or Ctrl+click on Windows/Linux) on any image in the editor opens it in the system default viewer via `openPath`.

Implementation details:
- `useEditorImageActivation` hook attaches capture-phase `click` and `mousedown` listeners to the editor container. Capture phase is required to intercept before BlockNote's own handlers.
- `mousedown` with the modifier key calls `preventDefault` to suppress BlockNote's focus/selection changes on the target image.
- Path resolution handles both direct `` clicks and clicks on BlockNote wrapper elements (resize handles, the visual media wrapper) by walking up to the nearest image ancestor via `IMAGE_ANCESTOR_SELECTOR`.
- Both `asset://localhost/` and `http://asset.localhost/` URL schemes are recognised and decoded to absolute filesystem paths.
- The Tauri `opener:allow-open-path` capability is scoped to `$HOME/**` — broad enough to cover any vault location under the user's home directory without granting unrestricted filesystem access. (The `opener:default` permission only covers URLs; `open_path` requires an explicit declaration.)
- Full test suite covers: both asset URL schemes, Cmd+click, Ctrl+click, plain click (no-op), non-image elements, images without asset-scheme src, clicks on BlockNote resize handles, and clicks on images outside the editor container.

## Test plan

- [ ] Open a note with images stored as `attachments/img.jpg`, `/attachments/img.jpg`, `../attachments/img.jpg`, and a filename with spaces — all should render without broken-image icons
- [ ] Cmd+click a high-resolution image in the editor — should open in Preview (macOS) at full native resolution for zooming
- [ ] Ctrl+click should do the same on Windows/Linux
- [ ] Plain click on an image should not open anything
- [ ] Cold-start the app (no prior `reload_vault` call) — images should load immediately without an asset-protocol error in the console
- [ ] `pnpm test` — all unit tests pass

### DIFF
--- a/src-tauri/capabilities/default.json
+++ b/src-tauri/capabilities/default.json
@@ -17,6 +17,10 @@
     "dialog:default",
     "updater:default",
     "process:default",
-    "opener:default"
+    "opener:default",
+    {
+      "identifier": "opener:allow-open-path",
+      "allow": [{ "path": "$HOME/**" }]
+    }
   ]
 }
--- a/src-tauri/src/commands/vault.rs
+++ b/src-tauri/src/commands/vault.rs
@@ -229,7 +229,7 @@ mod tests {
             .output()
             .unwrap();
 
-        let entries = list_vault(vault_path.into()).unwrap();
+        let entries = crate::vault::scan_vault_cached(vault_path).unwrap();
         assert!(!entries[0].archived);
 
         std::fs::write(
--- a/src-tauri/src/commands/vault/file_cmds.rs
+++ b/src-tauri/src/commands/vault/file_cmds.rs
@@ -160,8 +160,14 @@ pub fn copy_image_to_vault(vault_path: PathBuf, source_path: PathBuf) -> Result<
 }
 
 #[tauri::command]
-pub fn list_vault(path: PathBuf) -> Result<Vec<VaultEntry>, String> {
-    with_expanded_vault_root(path.as_path(), vault::scan_vault_cached)
+pub fn list_vault(
+    app_handle: tauri::AppHandle,
+    path: PathBuf,
+) -> Result<Vec<VaultEntry>, String> {
+    with_expanded_vault_root(path.as_path(), |expanded| {
+        crate::sync_vault_asset_scope(&app_handle, expanded)?;
+        vault::scan_vault_cached(expanded)
+    })
 }
 
 #[tauri::command]
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -161,9 +161,23 @@ fn setup_app(app: &mut tauri::App) -> Result<(), Box<dyn std::error::Error>> {
     {
         run_startup_tasks();
         spawn_ws_bridge(app);
+        prime_asset_scope_for_last_vault(app);
     }
 
     Ok(())
+}
+
+#[cfg(desktop)]
+fn prime_asset_scope_for_last_vault(app: &tauri::App) {
+    let Some(vault_path) = settings::get_last_vault() else { return };
+    let expanded = commands::expand_tilde(&vault_path).into_owned();
+    let path = Path::new(&expanded);
+    if !path.is_dir() {
+        return;
+    }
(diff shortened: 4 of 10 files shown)


nshikov:
@LucaRonin images attachment didn't work properly in all cases, so I added these fixes + added possibility to open image in default viewer on cmd + click

LucaRonin:
thank you! Fixes should be in the latest stable already, and I'll add image viewer soon. Closing this for now but keeping it as a reference!

nshikov:
@LucaRonin checked latest release, but relative paths to images - with ./ and ../ in it still don't work
thanks!

nshikov:
@LucaRonin also paths with spaces in it don't work and when I tried %20 to replace spaces it still don't work e.g. IMAGE%202022-10-15%2015%3A47%3A07.jpg
my fixes should fix it all if you want to apply first commit

Issue #257, opened 2026-04-24
MCP connection not found in CC Cli session


marcyagu:
In my Mac, with my CC cli session, I had to move Tolaria MCP server settings from `~/.claude/mcp.json` to file `~/.claude.json`, under `mcpServers` key (copying the same settings written automatically by the app inside the first file). 
Without doing that Claude was not able to find MCP server at all while now it seems to work. 

Additionally, even if the tool is file based, so adding a note through agent could be really straightforward, I would suggest to add also additional tools inside the MCP server for note creation/update/organize and so on, to use at maximum the features exposed by Tolaria App (like template, view creation/update/usage, type handling). If I don't get wrong, actually the MCP can only be used to read the notes from an agents session but the entire handling of vault and CRUD for note is missing.

Thank you

LucaRonin:
The agent can actually do that! and it doesn't need the MCP — it can just create files from the terminal. Does it say that it can't?

marcyagu:
Ideally yes. Anyway, if I don't get wrong, CC can't use agents from different folders outside its session directory, so if we want to use the Tolaria agent in multiple projects, we need to copy paste the Tolaria agent file in our project folder? Or we just need to ask claude to add a note in a specific folder (relating to our vault) and he is able to use the agent?

The flow that I'm imagining is:
- I'm working on a project
- I want to add a note (with specific type or metadata) in my personal-notes vault
- asking claude to do it directly in session
- it should be able to redact, commit and eventually push the note

Lastly, still we need to look at the MCP connection issue mentioned in first lines

LucaRonin:
should be fixed in today’s stable version! let me know. closing for now

marcyagu:
Seems fine! ty.
Just to confirm, I'll have to copy the vault auto-generated agents file in all my project manually to satisfy my imagined flow requirements explained above right?

LucaRonin:
should be fixed in the latest stable!

PR #286, opened 2026-04-25
fix: spawn ws-bridge with active vault from vaults.json (not hardcoded ~/Laputa)

leepoweii:
## Problem

The Tauri host always passes `VAULT_PATH=$HOME/Laputa` (the author's personal vault name) to the bundled MCP `ws-bridge.js` child process, **regardless of the user's `active_vault` in `vaults.json`**. As a result every MCP read tool — `search_notes`, `open_note`, `get_note`, `get_vault_context` — returns `ENOENT` for any user whose vault isn't literally `~/Laputa`. The Tolaria UI itself works fine (it reads `vaults.json` directly), so the failure is silent: users see "vault works in UI" + "MCP broken in their AI client" and assume their MCP client is misconfigured.

This silently breaks the headline AI-agent integration on every install where `~/Laputa` doesn't exist — which is essentially every user but the author.

This is likely the root cause of #257 (`MCP connection not found in CC Cli session`).

## Reproduction

```bash
# 1. Tolaria is running with any vault other than ~/Laputa selected
$ cat ~/Library/Application\ Support/com.tolaria.app/vaults.json
{"active_vault": "/Users/<you>/Documents/my-vault", ...}

# 2. Inspect the spawned ws-bridge env:
$ ps eww $(pgrep -f "Tolaria.*ws-bridge.js") | tr ' ' '\n' | grep VAULT_PATH
VAULT_PATH=/Users/<you>/Laputa     ← ignores active_vault

# 3. Any MCP tool call confirms it:
$ echo '{"id":1,"tool":"search_notes","args":{"query":"x"}}' | websocat -n1 "ws://[::1]:9710"
{"id":1,"error":"ENOENT: no such file or directory, scandir '/Users/<you>/Laputa'"}

$ echo '{"id":2,"tool":"open_note","args":{"path":"any.md"}}' | websocat -n1 "ws://[::1]:9710"
{"id":2,"error":"ENOENT: no such file or directory, realpath '/Users/<you>/Laputa'"}
```

`refresh_vault` works (it doesn't actually need filesystem access), so health checks may falsely report MCP as "up".

## Root cause

`src-tauri/src/lib.rs` had two functions that hardcoded `~/Laputa` as the vault path instead of reading `vault_list::load_vault_list()`:

```rust
// before (lib.rs:62-64, 82-84)
let vault_path = dirs::home_dir()
    .map(|h| h.join("Laputa"))
    .unwrap_or_default();
```

PR #7 ("remove hardcoded paths") removed the equivalent hardcoding from the Rust note-write path, but missed these two startup callers.

## Fix

- New helper `current_active_vault_path()` reads `vault_list::load_vault_list()` and returns `Option<PathBuf>` of the active vault — `None` when no vault is configured or when the configured path no longer exists.
- `run_startup_tasks()` and `spawn_ws_bridge()` both use it.
- When no vault is configured, `spawn_ws_bridge()` now skips spawning entirely (with an info log) instead of starting a bridge that points at a non-existent path. This avoids a silent ENOENT every MCP call would hit otherwise.

Verified by spawning the same `ws-bridge.js` manually with `VAULT_PATH=<real vault>` — every tool call works (`get_vault_context` returns 52 notes from my pilot vault, `search_notes` matches CJK content, `open_note` returns frontmatter + body).

## Test

Added `current_active_vault_path_returns_none_when_no_vault_dir_exists` regression test in `lib.rs` tests module — exercises the new helper and asserts the contract that any returned path must be a real directory. Comment also documents the original bug for future maintainers.

## Out of scope (potential follow-ups)

- **Vault switching at runtime**: `vaults.json` can change while Tolaria is running, but ws-bridge env is set once at spawn time. A follow-up could kill + respawn ws-bridge when active vault changes.
- **JS-side default**: `mcp-server/ws-bridge.js` line 7301 still has `process.env.HOME + "/Laputa"` as a fallback. With this fix, that path is no longer reachable from Tolaria-spawned bridges, but anyone running `node ws-bridge.js` directly still hits it. Consider replacing the JS fallback with an explicit error.
- **Legacy `~/.config/com.laputa.app`**: `vault_list.rs` still falls back to this for older installs — out of scope here, but worth tracking.

## Notes

- Diff is 1 file, +59/-9 lines.
- No new dependencies.
- Both callers use the same helper for consistency.
- `current_active_vault_path()` is `#[cfg(desktop)]` to match the callers.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -56,15 +56,30 @@ fn log_startup_result(label: &str, result: Result<usize, String>) {
     }
 }
 
-/// Run startup housekeeping on the default vault (migrate legacy frontmatter, seed configs).
+/// Resolve the active vault path from `vaults.json`.
+///
+/// Returns `None` when no vault list exists, no `active_vault` is set, or the
+/// configured path no longer points at a directory (e.g. user moved/deleted
+/// the vault outside the app).
+#[cfg(desktop)]
+fn current_active_vault_path() -> Option<PathBuf> {
+    let list = vault_list::load_vault_list().ok()?;
+    let path = PathBuf::from(list.active_vault?);
+    if path.is_dir() {
+        Some(path)
+    } else {
+        None
+    }
+}
+
+/// Run startup housekeeping on the active vault (migrate legacy frontmatter, seed configs).
+///
+/// No-op when no active vault is selected.
 #[cfg(desktop)]
 fn run_startup_tasks() {
-    let vault_path = dirs::home_dir()
-        .map(|h| h.join("Laputa"))
-        .unwrap_or_default();
-    if !vault_path.is_dir() {
+    let Some(vault_path) = current_active_vault_path() else {
         return;
-    }
+    };
     let vp_str = vault_path.to_str().unwrap_or_default();
     log_startup_result(
         "Migrated is_a to type on startup",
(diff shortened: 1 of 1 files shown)


leepoweii:
Closing temporarily — want to do more thorough manual verification on my end before maintainer review. Will reopen once verified. Sorry for the noise.

PR #288, opened 2026-04-25
Enhance build-release-history-page with validations and logging

krataratha:
Added validation for file existence and payload structure, included timestamp in HTML output, and improved logging.


LucaRonin:
not sure we need this, and I just upgraded the release page anyway. Thank you anyway, closing for now!

PR #291, opened 2026-04-25
feat: enhance IME handling in InlineWikilinkInput

laogui:
## Problem

#<60554d0b> handled IME composition at the React-state layer (skipping `syncValueFromEditor` and keydown-driven business logic during composition). Three issues remain visible in long compositions like Pinyin / Wubi / Kana, where the IME holds an active session for many keystrokes:

1. **Selection sync bypasses `isComposing`.** `onKeyUp` / `onClick` / `onMouseUp` still call `syncSelectionRange` in `useInlineWikilinkSelection`, which runs `setSelectionRange` → re-render → `useLayoutEffect` → `selection.removeAllRanges()` + `addRange()`. That tears down the IME composition context on every keystroke, producing scrambled previews mixing pinyin letters, committed glyphs, and the temporary IME preview.
2. **Stray text nodes after `compositionend`.** WebKit/Blink commit the final glyphs as a sibling text node next to the React-rendered `<span>`. `serializeInlineNode` picks them up so React state ends up with the right value, but the orphan node is never removed by reconciliation. Visible result: every committed character is duplicated.
3. **Focus theft on flush.** If the user moves focus away mid-composition (e.g. clicks Send or a sidebar item), `compositionend` fires on the original element and the post-flush focus restore steals the focus back.

## Solution

- Plumb `isComposingRef` into `useInlineWikilinkSelection` and short-circuit both `syncSelectionRange` and the layout-effect selection application during composition.
- On `compositionend` flush, force a remount of the contenteditable subtree (via a `renderVersion` bump) so any IME-injected stray text nodes are dropped along with the old DOM.
- Capture the target selection in a ref and restore it in a layout effect after the remount, but only refocus when the editor still owned `document.activeElement` at flush time, to avoid stealing focus.

## Tests

New regressions in `src/components/WikilinkChatInput.test.tsx`:
- `does not reset the DOM selection while IME composition is active`
- `clears IME-injected stray text nodes after composition ends`
- `does not steal focus back if it was moved elsewhere during composition end`

Each test was reverse-verified by undoing the corresponding fix and confirming it failed. Full vitest suite green.

## Why this didn't show up before

The existing test suite asserts callback invocations (React semantics) and runs in jsdom, which doesn't simulate the two real-browser side effects above (the browser-driven `Selection.collapse()` and the orphan text node). The new regression tests inject those side effects manually.

## QA

Manual native QA on macOS WKWebView: typed Pinyin into `AiPanel` and `CommandPaletteAiMode`, confirmed no scrambling, no duplication, and no focus theft when clicking sidebar / Send mid-composition.

### DIFF
--- a/src/components/InlineWikilinkInput.tsx
+++ b/src/components/InlineWikilinkInput.tsx
@@ -1,4 +1,5 @@
 import {
+  useLayoutEffect,
   useMemo,
   useRef,
   useState,
@@ -16,7 +17,11 @@ import {
   extractInlineWikilinkReferences,
   findActiveWikilinkQuery,
 } from './inlineWikilinkText'
-import { serializeInlineNode } from './inlineWikilinkDom'
+import {
+  readSelectionRange,
+  serializeInlineNode,
+  type InlineSelectionRange,
+} from './inlineWikilinkDom'
 import {
   buildPendingPasteState,
   type PendingPasteState,
@@ -145,6 +150,7 @@ export function InlineWikilinkInput({
     [entries, value],
   )
   const typeEntryMap = useMemo(() => buildTypeEntryMap(entries), [entries])
+  const isComposingRef = useRef(false)
   const {
     editorRef,
     selectionRange,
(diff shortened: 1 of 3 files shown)


LucaRonin:

PR #292, opened 2026-04-25
Fix/unicode title slug

laogui:
## Problem

Creating an untitled note (Cmd+N) and then setting its title to a CJK string (e.g. `你好`) wrote the file to disk as `.md` (stem-less), invisible in Finder and excluded from the in-app note list.

Root cause: `title_to_slug` in `src-tauri/src/vault/rename.rs` used `char::is_ascii_alphanumeric`, mapping every non-ASCII char to `-` and then filtering empty segments — `title_to_slug("你好")` returned `""`, so `expected_filename = ".md"`.

Trigger chain:
1. User Cmd+N → `untitled-note-<ts>.md` created.
2. User types CJK H1 in the editor.
3. Frontend `needsRenameOnSave` (Unicode-aware TS `slugify`) requests rename.
4. Rust `title_to_slug` collapses to empty → file renamed to `.md`.

## Solution

- Switch to `char::is_alphanumeric` (Unicode-aware), matching the frontend `slugify` in `src/hooks/useNoteCreation.ts` which already uses `\p{Letter}\p{Number}`.
- Fall back to `"untitled"` when no alphanumeric scalar survives, also matching the frontend behaviour for symbol-only titles.

## Tests

New regressions in `src-tauri/src/vault/rename.rs`:
- `test_title_to_slug_preserves_unicode_letters` (CJK, JP, mixed ASCII+CJK)
- `test_title_to_slug_falls_back_to_untitled_for_symbol_only_titles`
- `test_rename_note_with_cjk_title_writes_unicode_filename` — end-to-end: untitled note → CJK rename → asserts `你好.md` lands on disk and no stem-less `.md` exists.

`cargo test --lib vault::` → 341 passed.

## Out of scope (mentioned for awareness, not included)

`slugifyWikilinkTarget` (`src/utils/wikilink.ts`) and `slugifyPathStem` (`src/hooks/editorTabContent.ts`) still use the ASCII-only pattern. They don't produce stem-less files, but CJK-titled notes will not resolve as wikilink targets and will miss the `isUntitledRenameTransition` fast-path. Happy to follow up in a separate PR if maintainers prefer that direction.

### DIFF
--- a/src-tauri/src/vault/rename.rs
+++ b/src-tauri/src/vault/rename.rs
@@ -55,17 +55,25 @@ struct WikilinkUpdateSummary {
     failed_updates: usize,
 }
 
-/// Convert a title to a filename slug (lowercase, hyphens, no special chars).
+/// Convert a title to a filename slug (lowercase, hyphens, Unicode letters/digits preserved).
+///
+/// Falls back to `"untitled"` when the title contains no alphanumeric characters
+/// (matches the frontend `slugify` in `src/hooks/useNoteCreation.ts`).
 pub(super) fn title_to_slug(title: &str) -> String {
-    title
+    let joined = title
         .to_lowercase()
         .chars()
-        .map(|c| if c.is_ascii_alphanumeric() { c } else { '-' })
+        .map(|c| if c.is_alphanumeric() { c } else { '-' })
         .collect::<String>()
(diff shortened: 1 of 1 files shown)


LucaRonin:

PR #308, opened 2026-04-25
fix: insert file/folder paths dropped from Finder into AI chat input

### LINKED ISSUE #302: unable to drag folder path into AI chat
On macOS, it is common for me to open Finder, navigate to a directory, and then click and drag that folder into a Terminal window or a Claude Code session in order to copy the path to that directory into the working window. When trying to do this with the AI chat, it does not copy the path.

Steps to reproduce: 
1. Open AI Panel in Tolaria
2. Open Finder
3. navigate to a folder
4. click and drag the folder into the input window on the AI panel

Expected Result:
* the path to the selected folder if pasted into the window
To see expected behavior you can follow the same steps into a new Terminal window.

Actual Result:
* nothing gets pasted in the window

chrisvaughn:
Dragging a file or folder from macOS Finder into the AI chat composer
  now inserts its filesystem path as text. Enables Tauri's native drag-drop
  event system (dragDropEnabled: true) to access real filesystem paths, adds
  a tauri://drag-drop listener to the AI input with position-checking so
  drops only apply to the input, and fixes a broken type-discriminant check
  in useImageDrop that was silently swallowing all Tauri drop events.

  Fixes #302


LucaRonin:

PR #316, opened 2026-04-26
fix: AI agent CLIs reliably surface errors and accept tilde vault paths

jwm-axoni:
## Summary

Fixes a cluster of bugs that caused Tolaria's AI chat panel to silently fail with `"Claude Code finished without returning a reply."` regardless of what was actually wrong upstream.

### Root cause

When a user's `vaults.json` stored a vault path with a literal `~` (e.g. `~/Vaults/work-command-center`), the `stream_claude_agent` / `stream_ai_agent` Tauri commands passed it raw into `Command::current_dir(...)`. Rust's `Command` does not expand `~`, so spawn failed with `ENOENT`. The error then got swallowed because the frontend's `onDone` handler ran after `onError` and overwrote the error message with the empty-reply fallback. Net effect: every Claude/Codex run looked like it returned nothing, with no clue why.

This affected anyone whose vaults entry was added with a tilde — including the default in `vaults.json` for some setups.

### Fixes (in commit order)

1. **`chore: ignore .claude worktree paths in eslint`** — Claude Code worktrees live under `.claude/worktrees/<name>/` inside the repo and carry tracked `src-tauri/gen/` assets that the global `src-tauri/gen/` ignore (resolved relative to repo root) doesn't match. Adds `.claude/` to `globalIgnores` in `eslint.config.js` so `pnpm lint` succeeds when worktrees are present. Unblocks the lint gate for anyone using Claude Code worktrees during development.

2. **`fix: parse ISO date strings as local-timezone calendar dates`** — `formatDateValue('2026-03-31')` returned `'Mar 30, 2026'` for users west of UTC because `new Date('2026-03-31')` parses as UTC midnight. `resolveDateFromValue` now constructs date-only ISO inputs from local components. ISO datetimes with explicit time keep the existing parse path. Two pre-existing test-suite assertions (`propertyTypes.test.ts`, `DynamicPropertiesPanel.test.tsx`) start passing in EDT after this.

3. **`fix: surface real CLI errors instead of empty-reply fallback`** — Three coordinated changes so genuine CLI failures (rate limits, billing, model mismatch, auth, subprocess spawn) reach the chat panel:
   - `claude_cli.rs`: emit `Error` when the result event has `is_error: true`; fall back to `api_error_status` when the result text is empty.
   - `ai_agents.rs`: dispatch Codex `error` and `turn.failed` JSON events to `Error`.
   - `aiAgentStreamCallbacks.ts`: ignore trailing `Done` after `Error` so the error message is not overwritten.

4. **`fix: expand tilde in AI agent vault_path before subprocess spawn`** — The actual root cause. Both `run_agent_stream` (claude_cli.rs) and `run_ai_agent_stream` (ai_agents.rs) now normalize `vault_path` through `crate::commands::expand_tilde` before constructing the subprocess. Codex's run path inherits the same normalization since it reuses `request.vault_path` for cwd and MCP env.

5. **`fix: hide AI chat placeholder when editor is focused`** — The placeholder overlay in `InlineWikilinkEditorField` (`"Ask Claude Code about this note..."`) was an absolutely-positioned div whose only hide condition was a non-empty value. Clicking into the field with no text left the placeholder sitting on top of the caret. Adds `group` to the wrapper and `group-focus-within:hidden` to the overlay so the placeholder disappears the moment the contentEditable gets focus, while still re-appearing on blur.

## Verification

All five commits passed the local pre-commit gate (3243 vitest passing on the final commit). Focused tests added for each fix:

- Rust: `dispatch_event_handles_error_result`, `dispatch_event_error_result_falls_back_to_status`, `dispatch_codex_error_event_maps_to_error`, `dispatch_codex_turn_failed_event_maps_to_error`, `normalize_agent_request_expands_tilde_vault_path`, `normalize_agent_request_leaves_absolute_paths_intact`, `normalize_ai_agent_request_expands_tilde_vault_path`, `normalize_ai_agent_request_leaves_absolute_paths_intact`.
- TypeScript: new `aiAgentStreamCallbacks.test.ts` case for trailing-Done-after-Error; new `WikilinkChatInput.test.tsx` case for placeholder overlay class structure.

## Test plan

- [ ] Set a vault path with `~` (e.g. `~/Vaults/foo`) in `vaults.json` and confirm Claude Code chat returns a real response instead of the empty-reply message.
- [ ] Force a Claude rate-limit (e.g. with `ANTHROPIC_API_KEY` on a depleted account) and confirm the actual rate-limit error appears in chat instead of the empty-reply message.
- [ ] Click into the AI chat input on an empty draft and confirm the placeholder disappears immediately.
- [ ] Open a date property in EDT and confirm `2026-03-31` renders as `Mar 31, 2026`.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -6,7 +6,7 @@ import tseslint from 'typescript-eslint'
 import { defineConfig, globalIgnores } from 'eslint/config'
 
 export default defineConfig([
-  globalIgnores(['dist', 'coverage', 'src-tauri/resources/', 'src-tauri/target/', 'src-tauri/gen/', 'tools/']),
+  globalIgnores(['dist', 'coverage', 'src-tauri/resources/', 'src-tauri/target/', 'src-tauri/gen/', 'tools/', '.claude/']),
   {
     files: ['**/*.{ts,tsx}'],
     extends: [
--- a/src-tauri/src/ai_agents.rs
+++ b/src-tauri/src/ai_agents.rs
@@ -287,6 +287,20 @@ where
                 });
             }
         }
+        "error" => {
+            if let Some(message) = json["message"].as_str() {
+                emit(AiAgentStreamEvent::Error {
+                    message: message.to_string(),
+                });
+            }
+        }
+        "turn.failed" => {
+            if let Some(message) = json["error"]["message"].as_str() {
+                emit(AiAgentStreamEvent::Error {
+                    message: message.to_string(),
+                });
+            }
+        }
         "item.started" => emit_codex_item_event(json, false, emit),
         "item.completed" => emit_codex_item_event(json, true, emit),
         _ => {}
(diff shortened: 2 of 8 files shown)


LucaRonin:
this should be probably fixed by now! let me know if you still see it

PR #318, opened 2026-04-26
fix: expand tilde in vault cwd and guard agent stream edge cases

joceqo:
## Summary

- **Tilde expansion in vault `cwd`**: macOS does not expand `~` in `Command::current_dir` — paths like `~/Vaults/myvault` cause an ENOENT spawn failure. Fixed by calling `expand_tilde()` before passing the vault path to `current_dir`.
- **Suppress empty agent reply**: Tool-only responses (agent used tools but emitted no text) were showing "Claude Code finished without returning a reply." Fixed by returning an empty string when `actions.length > 0`.
- **`onDone` guard**: After `onError` finalises a message (sets `isStreaming: false`), `onDone` no longer overwrites it with the fallback text.
- **WebSocket CONNECTING guard**: `useAiActivity` was calling `ws.close()` on a socket still in `CONNECTING` state, producing a console warning. Now closes only when `readyState === OPEN`; also closes immediately in `onopen` if the component unmounted during connection.
- **Locale-safe test assertions**: Word-count assertions in `SearchPanel.test.tsx` and `noteListHelpers.extra.test.ts` now handle non-ASCII thousands separators (e.g. French locale uses narrow no-break space U+202F).

## Root cause

The tilde bug was the primary issue: vault paths starting with `~/` are stored verbatim and passed directly to Rust's `Command::current_dir`. The OS does not expand tildes in this context — only shells do. Claude CLI found the binary fine (`bin_exists=true`) but the `chdir` failed silently, causing every agent interaction to error out.

## Test plan

- [x] All 3190 frontend unit tests pass (`pnpm test`)
- [x] All 613 Rust tests pass (`cargo test`)
- [x] Agent interaction in the native app no longer shows "Claude Code finished without returning a reply." when vault path contains `~`
- [x] No "WebSocket closed before connection established" console warning on agent panel open/close

### DIFF
--- a/src-tauri/src/claude_cli.rs
+++ b/src-tauri/src/claude_cli.rs
@@ -290,6 +290,9 @@ struct StreamState {
     tool_inputs: HashMap<String, String>,
     /// The tool_use id of the block currently being streamed.
     current_tool_id: Option<String>,
+    /// Whether any TextDelta has been emitted — used to decide if the
+    /// final `result` text should be surfaced as a fallback.
+    text_was_emitted: bool,
 }
 
 /// Core subprocess runner shared by chat and agent modes.
@@ -309,7 +312,7 @@ where
         .stdout(Stdio::piped())
         .stderr(Stdio::piped());
     if let Some(dir) = cwd {
-        cmd.current_dir(dir);
+        cmd.current_dir(crate::commands::expand_tilde(dir).as_ref());
     }
     let mut child = cmd
         .spawn()
@@ -322,6 +325,7 @@ where
         session_id: String::new(),
         tool_inputs: HashMap::new(),
         current_tool_id: None,
+        text_was_emitted: false,
     };
 
     for line in reader.lines() {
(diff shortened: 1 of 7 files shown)


Uminily:
Hey @joceqo, the tilde-in-`current_dir` part of this PR is the root cause behind the freshly opened #486 (and matches what users on macOS / fish shell are reporting there). To make that specific fix easier to merge while this PR is `CONFLICTING`, I opened a focused, single-concern variant at #496 with unit tests covering the tilde and absolute-path cases, just for that one bullet of yours.

joceqo:
ok nice if we can have this issue fix in tolaria, if i remember well https://github.com/refactoringhq/tolaria/pull/316 fixed the issue bit better than this pr, i didnt had time to look at yours

LucaRonin:
I think this should be fixed by now! Thank you anyway!

joceqo:
thank you for you work Luca, huge improvements recently 💪

PR #335, opened 2026-04-26
fix: prevent IME stray text node accumulation in AI composer

suwonyoo:
## Summary

Korean / Japanese / Chinese typing in the AI Agent panel currently accumulates each composition's intermediate state instead of replacing it. Typing `안녕` repeatedly produces output like:

```
안녕ㅇ안녕ㅇ안녕ㅇ안녀ㅇ안ㄴㅇ안ㅇ안ㅇ아ㅇㅇㅇㅇ아앙아ㅇ
```

## Root cause

`InlineWikilinkInput` already routes IME composition through `isComposingRef` and forces a remount of the editor (`key={renderVersion}`) when a composition commits. That mostly works, but the IME inserts plain text nodes directly into the contentEditable div, **outside React's reconciliation tree**. The trailing-jamo siblings survive React's normal child diff and pile up alongside the React-rendered `<span>` segments — every subsequent composition serializes the previous strays back into the value.

Existing IME tests cover only a single composition cycle, where the `key`-based remount cleans the DOM. Across consecutive compositions in WKWebView (Tauri), the strays leak.

## Fix

`useInlineWikilinkSelection` already re-runs its layout effect on every value/selection change (and skips while `isComposingRef.current === true`). The fix removes any stray text node siblings of React-rendered children at that point, while preserving the trailing ZWSP that React itself emits as a caret anchor.

A safety guard skips cleanup when the editor holds **only** text (no spans/chips). This avoids blanking the field in test or programmatic flows that set `editor.textContent` directly before React has had a chance to mount its span.

## Test coverage

- New `WikilinkChatInput.consecutive-ime.test.tsx`:
  - `does not duplicate syllables across two consecutive compositions`
  - `does not duplicate across three consecutive compositions`
- Hardened `does not hijack Enter while IME composition is active` to query the live editor element rather than a stale, detached reference.

All 3269 frontend unit tests pass.

## Manual verification

- Built locally and replaced `/Applications/Tolaria.app`. Typing 안녕 repeatedly in the AI Agent panel now commits the correct text without per-syllable duplication.

## Test plan

- [ ] CI green
- [ ] Type Korean / Japanese / Chinese into the AI composer, including:
  - [ ] Single syllable then commit (existing behaviour)
  - [ ] Several consecutive syllables back-to-back
  - [ ] Mid-composition cursor moves / clicking elsewhere
- [ ] Plain ASCII typing still commits correctly
- [ ] Wikilink suggestion menu still triggers on `[[`

### DIFF
--- a/src/components/WikilinkChatInput.consecutive-ime.test.tsx
+++ b/src/components/WikilinkChatInput.consecutive-ime.test.tsx
@@ -1,0 +1,86 @@
+import { useState } from 'react'
+import { describe, it, expect, vi } from 'vitest'
+import { render, screen, fireEvent, waitFor } from '@testing-library/react'
+import { WikilinkChatInput } from './WikilinkChatInput'
+import type { VaultEntry } from '../types'
+
+const entries: VaultEntry[] = []
+
+function Controlled({
+  onDraftChange,
+}: {
+  onDraftChange?: (value: string) => void
+}) {
+  const [value, setValue] = useState('')
+  return (
+    <WikilinkChatInput
+      entries={entries}
+      value={value}
+      onChange={(next) => {
+        onDraftChange?.(next)
+        setValue(next)
+      }}
+      onSend={vi.fn()}
+    />
+  )
+}
+
+describe('WikilinkChatInput — consecutive IME compositions', () => {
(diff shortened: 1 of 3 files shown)


LucaRonin:

PR #355, opened 2026-04-27
feat: open inline images in a lightbox on double-click

yusoufu:
## Problem
Inline images inserted into a note can only be viewed at the size BlockNote renders them at — there is no way to see a full-size preview without opening the file outside Tolaria.

## Solution
Double-clicking any `` inside the editor opens it in a centered shadcn Dialog at up to 90vw / 90vh. Esc, backdrop click, or the overlay close it. Single-click is left alone so BlockNote selection / cursor placement keep working as before.

A shared `dblclick` listener on the editor container picks up image targets via `getDoubleClickedImageSrc`, which also filters tracking pixels (<16 px naturalWidth/Height) so empty inline GIFs don't open a blank viewer.

## Changes
- `src/utils/imageLightboxTarget.ts` — pure helper extracting an image src from an event target, with a tracking-pixel guard. Five unit tests.
- `src/hooks/useImageLightbox.ts` — wires `dblclick` on a container ref to open/close state.
- `src/components/ImageLightbox.tsx` — shadcn Dialog with a transparent backdrop showing the full-size image.
- `src/components/SingleEditorView.tsx` — three-line wire-up (hook + render).

## Test plan
- [x] `pnpm lint` — 0 errors
- [x] `pnpm exec tsc --noEmit`
- [x] `pnpm test` — 3217/3217 pass (5 new)
- [x] `pnpm playwright:smoke` — all green (ran in pre-push)
- [ ] Native Tauri QA (manual) — drop an image into a note, double-click, confirm lightbox; press Esc, click backdrop


LucaRonin:

PR #366, opened 2026-04-27
feat: add sidebar back/forward navigation arrows

d0lim:
## Problem

Tolaria already has full browser-style note back/forward navigation under the hood — `useNavigationHistory` (stack + cursor + dedupe + MAX=200 + skips deleted paths), `useAppNavigation` (auto-pushes on `activeTabPath` change), `useNavigationGestures` (mouse buttons 4/5), the `viewGoBack` / `viewGoForward` commands with `Cmd+←` / `Cmd+→` shortcuts, and matching menu items. But there's no **visible UI** for it. New users have no way to discover the feature, and even returning users can't tell at a glance whether back history is currently available.

## Solution

Add Back / Forward icon buttons to the top-left of the sidebar (`SidebarTitleBar`), wired to the same `handleGoBack` / `handleGoForward` that the keyboard shortcut, mouse buttons, and menu items already call. **No history-model changes** — this is purely an additional UI surface.

- New component `SidebarNavArrows` — two ghost icon buttons (`CaretLeft` / `CaretRight`, 24×24) wrapped in `ActionTooltip` that shows the existing platform-aware shortcut display (`⌘←` on macOS, `Ctrl+Left` elsewhere) via the existing `getAppCommandShortcutDisplay` helper.
- Mounted in `SidebarTitleBar` alongside the collapse button (`justify-end` → `justify-between`). Hidden when sidebar is collapsed (the title bar itself isn't rendered then) — keyboard / mouse buttons still work in that state.
- Buttons disable when the corresponding `canGoBack` / `canGoForward` is `false` (browser/Finder convention).
- Props drilled through `Sidebar` from the existing `useAppNavigation` destructure in `App.tsx`. No new global state, no Context, no new hooks.

### One small adjacent fix

The existing collapse button in `SidebarTitleBar` was missing `data-no-drag`, which means the surrounding window-drag region could swallow its click in some focus states. Added it while editing the same function (matches the new arrow buttons, which need `data-no-drag` for the same reason). Trivial defensive correctness, not a behavioural change.

## Tests

- New `SidebarNavArrows.test.tsx` — 7 behavioural tests: render with English / Simplified Chinese aria-labels, click → handler called, `disabled=true` → handler **not** called (for both directions), `data-no-drag` present.
- New `SidebarTitleBar.test.tsx` — 4 regression tests: arrows + collapse render together, both forward correctly, collapse still works, collapse hidden when `onCollapse` undefined while arrows still show.
- Full suite green: 279 files / 3333 tests, lint clean, `tsc --build --force` clean.

## Out of scope (intentionally not included)

- Tracking `SidebarSelection` changes (filter / folder clicks) in history. Current model tracks only `activeTabPath`; expanding that is a larger UX call worth deferring until the buttons are actually in use.
- Unifying `useNavigationHistory` with the separate `neighborhoodHistory` (Escape-out-of-note) stack.
- Persisting history across app restarts.
- Putting the arrows in `LinuxTitlebar` or in the `NoteList` header when the sidebar is collapsed — sidebar-collapsed users keep the keyboard shortcut and mouse buttons.

## Screenshots

- You can see arrows at sidebar.
스크린샷 2026-04-28 오전 1 24 04


LucaRonin:
not sure this is the right direction! I tried it locally and doesn't feel 100% right — will think more about it. Closing for now

thank you for submitting this anyway!

PR #371, opened 2026-04-27
fix: cursor pointer on clickable elements

glhrmv:
Adds `cursor: pointer` on many elements of the UI that one would reasonably expect the cursor to turn into a pointer when hovering over.


LucaRonin:

PR #376, opened 2026-04-28
feat: Add note list context menu

the-jwoo:
## Summary
- add a right-click context menu for note rows with note, retargeting, file, archive, and delete actions
- wire row-specific note actions from App into NoteList while preserving the existing Changes view discard/restore menu
- add regression coverage for visible menu actions, row-specific callbacks, and stateful labels

image

## Tests
- pnpm test src/components/NoteList.context-menu.test.tsx


LucaRonin:
this is conflicting and needed some rework but it's nice! will credit you as co-author when it goes live later today

PR #392, opened 2026-04-28
fix: wire Windows menu events and gate macOS-only menu items

bloknayrb:
On Windows/Linux, Tauri attaches menus to WebviewWindow rather than App, so App::on_menu_event never fires for menu clicks. Register WebviewWindow::on_menu_event for non-macOS builds so clicks reach the frontend event pipeline.

Gate services() and show_all() behind #[cfg(target_os = "macos")] — they rendered as visible but non-functional items on Windows. Gate hide() and hide_others() behind #[cfg(not(target_os = "linux"))] per Tauri 2 platform availability.

Export MAIN_WINDOW_LABEL from window_state so menu.rs can reference the window by a typed constant instead of a bare string literal.

### DIFF
--- a/src-tauri/src/menu.rs
+++ b/src-tauri/src/menu.rs
@@ -1,3 +1,7 @@
+#[cfg(not(target_os = "macos"))]
+use crate::window_state::MAIN_WINDOW_LABEL;
+#[cfg(not(target_os = "macos"))]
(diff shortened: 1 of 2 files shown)


bloknayrb:
Note: all edits made in this and other PRs were by Claude code.

PR #412, opened 2026-04-28
fix: vault watcher reload loop on iCloud .git symlink vaults

kossoy:
## Summary

- Stop the watcher firing on `.git -> .git.nosync` paths, which was triggering a `reloadVault` loop on iCloud-backed vaults — the "Reload vault" status badge flickered every ~1s while idle. Resolves the symlink (and `gitdir:` worktree pointers) at watcher start and rejects any path inside the resolved directory. Also filters `.gitstatus.*` cache dirs and `*.icloud` placeholders.
- Fix `scripts/run-vitest-coverage.mjs` for standalone pnpm installs: it was passing the pnpm Mach-O / ELF binary to `process.execPath` (node) and crashing the pre-push hook with `SyntaxError: Invalid or unexpected token`. Now only takes that fast path when `npm_execpath` actually looks like a JS module.

## Repro for the watcher fix

iCloud / Dropbox Obsidian vaults commonly use the `.git -> .git.nosync` trick (see vault `.gitignore`). Background `git fetch` calls (auto-pull, `git_remote_status`, gitstatusd) write to the real `.git.nosync/` directory, which the existing `.git` component filter does not match. Each event fired `vault-changed` → `reloadVault`, with a 350ms debounce that produced multiple concurrent vault scans per second (visible as `Skipped replacing cache because another writer is active` warnings in the Rust log).

## Test plan

- [x] `cargo test --lib vault_watcher` — 7 tests, all green (3 new resolver cases + 1 new ignore case + extended temp-file coverage)
- [x] Reproduced loop on real iCloud vault: ~40 cache-skip lines/min before, 0 in 60s after
- [x] `pnpm test:coverage` — 3378 tests pass with the script fix on a standalone pnpm install


kent1976 (commented):
App

PR #434, opened 2026-04-29
Add default frontmatter properties for types

IridiumWasTaken:
**Problem statement**

For each type, different properties per entry ("note") may be suitable and helpful. At the moment, for each note belonging to a specific type, one has to add the properties manually to each entry.
I thought it would be helpful if there was a way to have each entry automatically have specific properties with optional default values for these properties.

I implemented it this way:
- One can now add default frontmatter properties to types the same way one can add a markdown template.
- Default frontmatter properties can have an optional default value.
- When a new note of that type is created, the properties are automatically added.

https://github.com/user-attachments/assets/84bf6d99-3737-4e51-94cc-94bab76d64dd

**Tests run**

I ran both the pre-commit and pre-push tests.
I tested functionality on Mac OS 26.4.1, Apple Silicon.


glhrmv:
I really like this idea but I think it could live in a better location than the "customise" pop-up. Perhaps the Properties pane for the type? Since that is conceptually akin to a "class definition", default values for fields should live there.

IridiumWasTaken:
> I really like this idea but I think it could live in a better location than the "customise" pop-up. Perhaps the Properties pane for the type? Since that is conceptually akin to a "class definition", default values for fields should live there.

I agree somewhat, but since the markdown template already lives in the customise pop-up, I thought it would be closest to the current version if properties lived there too. Maybe both markdown template and default frontmatter properties should be defined somewhere else.

LucaRonin:
This is now doable in the properties panel! You can create empty properties there that propagate as placeholders to type instances, or assign a default value in the type file that will propagate as well!

Issue #441, opened 2026-04-29
Language selection does not persist across app restarts


BrunoMiguelMonteiro:
It seems that the language  selected is reseted on app relaunch.

### Steps to reproduce:

- Open the app
- Navigate to Settings
- Select "Português" as the display language
- Save the setting — the UI language updates correctly
- Quit and relaunch the app

### Expected behavior: 
The app launches in Português, as previously selected.

### Actual behavior: 
The app reverts to English (the default) on every restart. The language preference is not persisted between sessions.

### Additional context: 
The setting appears to apply correctly within the current session, which suggests the language switching logic itself works. The issue is likely in how the preference is written to (or read from) persistent storage — the value may not be saved to disk, or the saved value may not be read on startup.

---

Tolaria sanitized diagnostics
Generated: 2026-04-29T21:28:37.568Z
Build: 2026.4.29
Release channel: stable
Runtime: tauri
Platform: MacIntel
User agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)

Recent diagnostics:
No safe recent diagnostics were available.

Notes: paths and token-like strings are redacted. This bundle is sanitized and optional.

LucaRonin:
Should be fixed in today's stable release! Let me know if you still see it — closing for now.

PR #447, opened 2026-04-30
fix: persist UI language setting across app restarts

IISweetHeartII:
## Problem

`normalize_ui_language()` in `settings.rs` only recognized `en` and zh-CN variants. Every other locale (`ko-KR`, `fr-FR`, `ja-JP`, `pt-BR`, etc.) was silently normalized to `None` and written as `null` to `settings.json`. On next launch the null was read back, causing the app to revert to English on every restart.

Reproduces #441.

## Fix

Extend the allowlist to every locale in the frontend's `APP_LOCALES` and return the canonical code unchanged. Also canonicalize legacy `zh-Hans` stored values to `zh-CN` for consistency with the frontend.

## Tests

- Renamed `test_invalid_ui_language_is_filtered` → `test_unsupported_ui_language_is_filtered` (uses `ar-SA` as an actually unsupported locale)
- Added `test_supported_ui_languages_are_preserved` covering all 13 non-English locales
- Updated zh-Hans → zh-CN in existing roundtrip tests

All 759 Rust tests pass.


LucaRonin:

PR #492, opened 2026-05-02
fix: hide formatting toolbar during IME composition

nelsonli33:
## Summary

- CJK IME users (注音/拼音/倉頡, Japanese, Korean) saw the floating formatting toolbar pop up during composition because BlockNote's transient selection flips `FormattingToolbarExtension`'s `show` state. This obscures the IME candidate window and breaks composition flow.
- New `useEditorComposing(editor)` hook tracks composition via `document`-level capture listeners filtered to the editor's contenteditable. Gates `TolariaFormattingToolbarController`'s `isOpen` on `!isComposing`.
- Mirrors the existing `useCompositionAwareEditorChange` pattern in `SingleEditorView.tsx` (which only protected `onChange`, not toolbar visibility).

## Implementation note

`useEffect` deps on `editor.domElement` (not `editor`) because BlockNote's ProseMirror view initializes asynchronously — `domElement` transitions from `undefined` to defined, and the listener must re-attach to the real contenteditable when that happens.

## Test plan

- [x] New unit test in `tolariaEditorFormatting.behavior.test.tsx` verifies the gate (composition → `open: false`, end → restored)
- [x] `pnpm test` — 3592 tests pass
- [x] `pnpm lint` + `npx tsc --noEmit` clean
- [x] `pnpm playwright:smoke` — 39 smoke tests pass
- [x] Manual QA on macOS WKWebView (Tauri dev) with 注音 IME — toolbar no longer fires mid-composition
- [x] Manual QA — text selection still shows toolbar normally (no regression)


LucaRonin:

LucaRonin:
Merged after some tweaks — closing this now.

PR #503, opened 2026-05-02
feat: vault provider abstraction with iCloud Drive support

Ancordss:
## Why

Tolaria is currently a desktop-only app. To enable future independent Apple clients (iPhone/iPad) that can open the same vault files, the app needs a way to manage vaults stored in iCloud Drive — not just local folders. Without this, there's no path to cross-device vault sharing that preserves Tolaria's files-first, offline-first model.

This PR lays the architectural foundation: a **vault provider abstraction** that separates "where files live" from "how notes work," so the same vault engine can operate over different storage backends. The first two providers are `local-folder` (existing behavior) and `icloud-drive` (new).

This also fixes a real user pain point: users who store vaults in iCloud Drive today have no visibility into sync state, no protection against cross-device overwrites, and no way for the app to behave differently for cloud-synced files.

## Summary

Introduces a **vault provider abstraction** that lets users explicitly choose how their vault is stored — **Local Folder** or **iCloud Drive** — and surfaces that choice throughout the UI.

### Key changes

- **Provider selection on vault creation and open**: A "Choose Vault Storage" dialog appears before the folder picker, letting users pick between Local Folder and iCloud Drive. When iCloud is selected, Finder opens directly to the iCloud Drive root.
- **iCloud status badge in status bar**: A cloud icon badge shows provider state (available/syncing/degraded/unavailable) alongside existing Git controls. iCloud and Git coexist independently.
- **Provider metadata persistence**: `providerType` and `providerRoot` are persisted per vault in `vaults.json` with backward-compatible migration of legacy entries.
- **Provider validation (Rust backend)**: A `validate_vault_provider_selection` command canonicalizes paths, detects iCloud Drive root on macOS, and validates explicit vs inferred provider choices.
- **Stale-save detection**: File identity (mtime + size) is captured when a note opens. Before saving on iCloud vaults, the identity is rechecked — if another device changed the file, a blocking recovery dialog offers "Reload from disk" or "Save as recovered copy."
- **Provider status runtime**: `currentStatus()` and `subscribeStatus()` adapter reports provider availability. Saves are blocked with a toast when provider is unavailable.
- **Bounded reconciliation**: iCloud vaults trigger a vault reload on window focus to catch files synced while the app was in the background.
- **Bug fix — remove vault**: Removing the last vault now properly clears `vaultPath` and redirects to the WelcomeScreen.

### Testing

- All **3620 Vitest unit tests** pass
- All **39 Playwright smoke tests** pass
- **Rust tests** pass for `vault_list` and `vault_provider`
- Frontend coverage >=70%, Rust coverage >=85%
- Clippy + rustfmt clean

### Design decisions

- **iCloud and Git coexist**: iCloud is a storage provider, Git is version control. They are independent.
- **Provider abstraction is narrow**: providers answer "where are the files" and "what's the sync state." Note semantics stay above the boundary.
- **Legacy migration is automatic**: existing entries default to `local-folder` on load.
- **Stale-save is iCloud-only for now**: local-folder vaults skip the identity check since there's no cross-device sync risk.


Ancordss:
## Next steps — Provider expansion roadmap
Now that the provider abstraction is in place, here’s the plan for evolving it:
### Short-term improvements to this PR’s UI
- **Replace the two-card selection with a dropdown**: Instead of showing Local Folder and iCloud Drive as separate cards, use a single dropdown/select component where the user picks their provider. This scales better as we add more options and keeps the UI clean.
- **Make Git a first-class provider**: Git should appear in the provider dropdown as the default option — not as a separate concept. The mental model becomes: “How do you want to sync this vault?” → Git (default) / iCloud Drive / Google Drive / Dropbox. Git is the default because it’s what Tolaria was built around and gives the most control.
### Provider dropdown options (target)

| Provider | Sync mechanism | Status |
| :--- | :--- | :--- |
| **Git (default)** | Git remotes (GitHub, GitLab, etc.) | Current behavior, should become the default dropdown option |
| **iCloud Drive** | Apple ecosystem automatic sync | Implemented in this PR |
| **Google Drive** | Cross-platform cloud sync | Next to implement — uses the same provider abstraction |
| **Dropbox** | Cross-platform cloud sync | Same pattern as Google Drive |



### Why Git as default provider makes sense
- Tolaria is git-first by design — version history, conflict resolution, and sync all work through Git today
- New users who pick Git get the full Tolaria experience out of the box
- Cloud providers (iCloud/GDrive/Dropbox) become alternative sync strategies for users who want simpler cross-device access without managing remotes
- The provider abstraction already supports this — Git would just be providerType: 'git' with full capability reporting
### Architecture notes for future providers
The provider boundary (src/lib/vaultProviders.ts, src-tauri/src/vault_provider.rs) is intentionally narrow. Adding Google Drive or Dropbox means:
1. Add the provider type to the enum
2. Add detection logic in Rust (similar to iCloud root detection)
3. Add a capability mapping
4. The rest (note engine, UI, save paths, stale-save detection) works unchanged
No core vault logic needs to change — that’s the point of the abstraction.

LucaRonin:
I understand the premise of this PR but I disagree that this is needed 🙂

Tolaria can already handle vaults that live in Dropbox / iCloud / Google Drive folders — you just need to have them physically in the file system. For mobile we will see, but for desktop none of this feels blocking.

Also a note about the fact that "iCloud and Git coexist" is absolutely not possible — git is not compatible with iCloud sync for several reasons.

PR #534, opened 2026-05-04
feat: enter neighbourhood view from any note title

glhrmv:
The "neighbourhood view" feature is really great but it could use more ways of triggering it. This MR triggers the neighbourhood view when `<Command>`-clicking any clickable note title - i.e. if it would otherwise bring you to the note itself when clicking on it, you can now `<Command>`-click on it to show the respective neighbours.


LucaRonin:
I agree, and I also think we need a button somewhere to trigger it, for people who don't know it's there (I suspect it's the majority right now). Need to think about it

glhrmv:
I think it's fine that it's not very discoverable for now. One suggestion that comes to mind (that would complement this PR) is adding a dedicated button for it somewhere in the editor panel or the properties panel - the former seems best but the editor panel is already quite cluttered as-is, and perhaps this is the kind of thing best left hidden from the UI. 

In the meantime, I will add some tests to this branch to address some of the concerns raised by the bots, later today. Regarding using `Context` to avoid drilling through so many layers: I agree in spirit, but I will not address this here, as it is too small a concern to use as launchpad into a big refactor.

LucaRonin:

LucaRonin:
Merged after some tweaks — closing this now.

PR #557, opened 2026-05-06
feat: render mermaid diagrams in markdown content

export-mike:
Adds a MermaidDiagram component that lazily imports `mermaid` and renders ```mermaid fenced blocks in MarkdownContent (used by AI messages and the editor preview).


LucaRonin:
This is already available! you can do /mermaid and create a new mermaid diagram

PR #692, opened 2026-05-17
Allow symlinks pointing outside the vault

alexbrasetvik:
Hi.

Thanks a lot for Tolaria. I'm experimenting with using it in a way where we federate knowledge bases from multiple different git repositories. To achieve that, at least for now, I'm currently symlinking across them. I opened [this feature request for federated/multi-vault](https://tolaria.canny.io/feature-requests/p/federated-multi-vault-support) for more first class support for that approach.

Tolaria currently rejects symlinks, since `canonicalize()` resolves them to being outside the vault root. I imagine that's a security feature, so MCP tools cannot reach outside the vault, which sounds pretty reasonable.

This PR allows symlinks inside the vault to point outside it, but rejects any attempt at using "..".

This was Claude assisted, and though I'm not experienced with the tech stack, it looks reasonable to me.


alexbrasetvik:
With native multi vault support, this workaround isn't needed.

PR #711, opened 2026-05-20
fix: include linuxbrew node path in MCP runtime fallbacks (#705)

### LINKED ISSUE #705: Tolaria desktop launcher cannot find node for Codex/MCP bridge
## Summary\nLaunching Tolaria from the desktop entry fails to find  when the app tries to start the Codex/MCP bridge. The UI shows:\n\n> Error: node not found in PATH or common install locations\n\n## Steps to Reproduce\n1. Launch Tolaria from the desktop entry / AppImage.\n2. Open the Codex agent flow.\n3. Trigger an action that starts the MCP bridge.\n\n## Expected Behavior\nTolaria should find  and start the Codex/MCP bridge successfully.\n\n## Actual Behavior\nThe bridge startup fails with:\n\n\n\n## Environment\n- Tolaria launched via \n- AppImage path: \n- Host shell has Node installed at \n- GUI launch does not appear to inherit the shell PATH\n\n## Notes\nThe current desktop launcher only sets  and does not prepend the Brew Node directory to . That makes the failure look like a launcher environment issue rather than a vault/content issue.\n\n## Suggested Fix\nEither:\n- set  in the desktop launcher before starting the AppImage, or\n- ship a small wrapper script that exports the required PATH and then launches Tolaria.\n\n

mvanhorn:
## Summary

Linux AppImage users with Linuxbrew can now start the Codex/MCP bridge from the desktop entry. `fallback_node_paths` in `src-tauri/src/mcp.rs` already covers macOS Homebrew, Windows ProgramFiles, and per-user nvm/mise/asdf/volta shims, but Linuxbrew (`/home/linuxbrew/.linuxbrew/bin/node`) was missing. When GUI launches don't inherit the shell PATH, the fallback list is the only safety net.

Adds the system-wide Linuxbrew path to `fallback_node_paths` and the per-user `~/.linuxbrew/bin/node` to `node_binary_candidates_for_home`. The existing `.is_file()` filter means machines without Linuxbrew see no behavior change.

The existing `node_binary_candidates_include_shell_managed_installs` test now also asserts the home-scoped Linuxbrew candidate.

Localization: no UI copy changes.

Fixes #705


mvanhorn:
Really appreciate the merge here too, @LucaRonin. The linuxbrew node path means AppImage + Linuxbrew users can finally start the MCP bridge from the desktop entry.

Issue #690, opened 2026-05-17
Could not start opencode in Windows


amr-swalha:
> Hello, issue still exists even after update
> 
> <img width="307" height="324" alt="Image" src="https://github.com/user-attachments/assets/1d3d3863-9c41-402b-a3df-f7cd9b1b28b7" /> 

 _Originally posted by @amr-swalha in [#604](https://github.com/refactoringhq/tolaria/issues/604#issuecomment-4471129525)_

LucaRonin:
not 100% what's the problem here, however updated Windows OpenCode discovery so PATH/shell lookup prefers real Windows CLI targets (.cmd/.exe/.bat/.com) over extensionless POSIX wrappers, matching the prior Claude/Codex fix pattern

going out today with the stable release!

LucaRonin:
Should be fixed in today's stable release! Let me know if you still see it — closing for now.

Issue #568, opened 2026-05-07
Error: Failed to spawn gemini: %1 is not a valid Win32 application. (os error 193)


DHS1001:
When trying to use Gemini CLI in the AI panel, it returns the below error message:
Error: Failed to spawn gemini: %1 is not a valid Win32 application. (os error 193)

<img width="493" height="243" alt="Image" src="https://github.com/user-attachments/assets/df1a9f25-8743-4953-afc7-e5a2a95af2be" />

I am running the latest version 2026.5.6 in Win 11. And Claude is woking as expected. Not sure if it is a bug?

Thanks

DHS1001:
Thanks 
It is working in the latest 2026.5.7.

PR #716, opened 2026-05-20
fix: prefer .cmd shim over POSIX wrapper for claude/codex on Windows (#695)

### LINKED ISSUE #695: [Windows] Error: Failed to spawn claude: %1 is not a valid Win32 application. (os error 193)
## Summary

Opening the AI chat panel and sending a message to **Claude** on Windows fails immediately with:

```
Error: Failed to spawn claude: %1 is not a valid Win32 application. (os error 193)
```

This is the same class of bug that was previously fixed for **Gemini** in #568 — but Claude (and Codex) still carry the original implementation that does not filter the `where claude` output for valid Windows-executable extensions.

## Environment

- OS: Windows 11 Pro 26200
- Tolaria: `main` @ `c1a5042` (latest)
- Claude CLI: `2.1.143 (Claude Code)`
- Installed via: `npm install -g @anthropic-ai/claude-code`
- Node: v26.0.0, npm: 11.12.1
- npm prefix: `C:\Users\<me>\AppData\Roaming\npm`

## Reproduction

1. On Windows, install Claude CLI via npm (`npm i -g @anthropic-ai/claude-code`). npm creates **two** shims side by side in the npm prefix dir:
   - `claude` (POSIX shell wrapper, **no extension**, intended for Git Bash / MSYS / WSL passthrough)
   - `claude.cmd` (Windows batch shim)
2. Open Tolaria, open the AI chat panel, select **Claude**, send any message.
3. The panel immediately shows: `Error: Failed to spawn claude: %1 is not a valid Win32 application. (os error 193)`.

Confirming what `where` returns on the machine:

```powershell
PS> where.exe claude
C:\Users\<me>\AppData\Roaming\npm\claude
C:\Users\<me>\AppData\Roaming\npm\claude.cmd
```

## Expected

Tolaria spawns `claude.cmd` (or the underlying `cli.js` via the existing `command_target_avoiding_windows_cmd_shim` codepath) and streams the response.

## Actual

Tolaria spawns the extensionless POSIX `claude` file. Windows then tries to load it as a PE32 image, the header is invalid, `CreateProcessW` returns `ERROR_BAD_EXE_FORMAT` (193), and Rust surfaces it as `%1 is not a valid Win32 application`.

## Root cause

In [`src-tauri/src/claude_cli.rs`](https://github.com/refactoringhq/tolaria/blob/main/src-tauri/src/claude_cli.rs#L120-L135), `path_from_successful_output` picks the **first existing line** from `where claude` without checking that the path actually has a Windows-executable extension:

```rust
fn path_from_successful_output(output: &std::process::Output) -> Option<PathBuf> {
    if !output.status.success() {
        return None;
    }

    String::from_utf8_lossy(&output.stdout)
        .lines()
        .find_map(|line| {
            let trimmed = line.trim();
            if trimmed.is_empty() {
                return None;
            }
            let candidate = PathBuf::from(trimmed);
            candidate.exists().then_some(candidate)
        })
}
```

`where claude` returns the POSIX wrapper first and the `.cmd` shim second, so this returns the wrong one. The subsequent `build_claude_command` → `command_target_avoiding_windows_cmd_shim` path only handles `.cmd`/`.bat` shims; an extensionless POSIX script falls through and gets spawned directly.

`find_executable_binary_candidate` (the well-known candidate list) **does** filter for `.bat/.cmd/.com/.exe` on Windows via `has_windows_cli_extension` — but it is only consulted *after* `find_claude_binary_on_path` succeeds, so it never runs in this scenario.

## Why Gemini is not affected

This was already fixed for Gemini in #568. [`src-tauri/src/gemini_discovery.rs`](https://github.com/refactoringhq/tolaria/blob/main/src-tauri/src/gemini_discovery.rs#L86-L97) filters by `has_windows_cli_extension` on Windows and even ships a regression test (`windows_path_lookup_prefers_cmd_shim_over_extensionless_npm_script`):

```rust
fn first_existing_path_for_platform(stdout: &str, windows: bool) -> Option<PathBuf> {
    let mut paths = stdout.lines().filter_map(existing_path);
    if windows {
        return paths.find(|path| crate::cli_agent_runtime::has_windows_cli_extension(path));
    }

    paths.next()
}
```

The same fix needs to be mirrored into Claude.

## Same bug also affects Codex

[`src-tauri/src/codex_cli.rs`](https://github.com/refactoringhq/tolaria/blob/main/src-tauri/src/codex_cli.rs#L87-L104) has the identical `path_from_successful_output` / `first_existing_path` pair without the Windows extension filter. Anyone with a duplicate Codex shim layout on Windows will hit the same error 193.

## Suggested fix

Mirror the Gemini implementation in `claude_cli.rs` and `codex_cli.rs`, ideally by promoting `first_existing_path_for_platform` into `cli_agent_runtime` so all three discovery sites share one implementation. Cover with the same regression test pattern as `windows_path_lookup_prefers_cmd_shim_over_extensionless_npm_script`.

Happy to send a PR if helpful.

mvanhorn:
## Summary

Tolaria on Windows now spawns the `.cmd` shim when launching Claude or Codex from the AI chat panel, instead of trying to load the extensionless POSIX wrapper as a PE32. The reporter traced this to `path_from_successful_output` in `claude_cli.rs` and `codex_cli.rs` picking the first existing `where` result without checking for a Windows-executable extension - matching the same class of bug `gemini_discovery.rs` already fixed via #568.

## Root Cause

> "`where claude` returns the POSIX wrapper first and the `.cmd` shim second, so this returns the wrong one. The subsequent `build_claude_command` -> `command_target_avoiding_windows_cmd_shim` path only handles `.cmd`/`.bat` shims; an extensionless POSIX script falls through and gets spawned directly." - @tieupham267, #695

On Windows, `npm install -g @anthropic-ai/claude-code` creates two side-by-side files in the npm prefix:
- `claude` - a POSIX shell wrapper (for Git Bash / MSYS passthrough)
- `claude.cmd` - the Windows batch shim

`where claude` returns both, POSIX wrapper first. The old `path_from_successful_output` selected by first-existing-line, which gave the POSIX wrapper. Windows then tried to load it as a PE32 image, `CreateProcessW` returned `ERROR_BAD_EXE_FORMAT (193)`, and the AI chat panel surfaced "%1 is not a valid Win32 application." The same class of bug for Gemini was fixed in #568 - `gemini_discovery.rs:90` already filters via the existing shared utility `cli_agent_runtime::has_windows_cli_extension`.

This PR mirrors that fix in `claude_cli.rs` and `codex_cli.rs`. Non-Windows behavior is unchanged - the same first-existing-line semantics apply via the new `first_existing_path_for_platform(stdout, false)` branch.

## Validation

- `cargo fmt --manifest-path src-tauri/Cargo.toml --check` passes
- `cargo build --manifest-path src-tauri/Cargo.toml` passes
- `cargo test --manifest-path src-tauri/Cargo.toml --lib claude_cli` - 49 passed, 0 failed
- `cargo test --manifest-path src-tauri/Cargo.toml --lib codex_cli` - 25 passed, 0 failed
- New regression test in each file: `windows_path_lookup_prefers_cmd_shim_over_extensionless_npm_script`, modeled on the existing test of the same name in `gemini_discovery.rs:219`.

## Follow-up

`opencode_discovery.rs:80` carries the same unfiltered `path_from_successful_output` pattern; that surfaces as issue #690 on Windows. Happy to send a follow-up PR mirroring this fix into opencode if you'd like.

Localization: no UI copy changes.

Fixes #695


mvanhorn:
Thanks @LucaRonin for merging the Windows .cmd shim fix. Good to have Claude/Codex launching from the AI chat panel without the PE32 error.

PR #778, opened 2026-05-30
Update mcp.rs, use timeout for find_on_path

HernandoR:
find_on_path calls external command, which may hang or takes really long time. add time out wrap to avoid infinite  waiting


HernandoR:
my case here is that i have a broken bashrc which looped and stuck. But my daily using shell is zsh. find_in_user_shell tries bash and therefore stuck forever

PR #800, opened 2026-06-02
fix: respect global git config instead of always setting vault@tolaria.md

### LINKED ISSUE #797: Git commits default to vault@tolaria.md, causing GitHub misattribution to another user
Hi, I noticed a potentially confusing/security-sensitive issue with Tolaria's Git sync.

When Tolaria initializes or syncs a vault using Git, it appears to set the repository-local Git identity to:

user.name=Tolaria
user.email=vault@tolaria.md

This caused commits in my private GitHub repository to be displayed as authored by another GitHub user, apparently because GitHub associates vault@tolaria.md with that account.

Image

Image

My global Git config was correctly set to my own GitHub identity, but the repository-local config overrode it.

After manually changing the local config:

git config user.name "My Name"
git config user.email "my-github-noreply-email@users.noreply.github.com"

new commits created through Tolaria were correctly attributed to me.

TolarIA should probably avoid using a default email address that may be associated with a real GitHub account.

### Environment
OS: Windows
Git remote: GitHub private repository
Git sync performed through Tolaria
Thanks!

yikf:
Fixes #797.

### Problem
Tolaria was always setting the Git user.email to vault@tolaria.md in the local repository config, even when the user already had a global Git configuration.

### Solution
Modified ensure_author_config to check for existing user.name and user.email config in all scopes (local, global, system) using git config --get, and only set the fallback defaults if no existing config is found.

### Testing
- Verified the fix works correctly with and without global Git config
- All tests pass (966 tests)

### DIFF
--- a/src-tauri/src/git/mod.rs
+++ b/src-tauri/src/git/mod.rs
@@ -287,17 +287,20 @@ fn run_git(dir: &Path, args: &[&str]) -> Result<(), String> {
     ))
 }
 
-/// Set local user.name and user.email if not already configured.
+/// Set local user.name and user.email only if no author config exists at any scope.
 pub(crate) fn ensure_author_config(dir: &Path) -> Result<(), String> {
     for (key, fallback) in [("user.name", "Tolaria"), ("user.email", "vault@tolaria.md")] {
-        let local = git_command()
-            .args(["config", "--local", key])
+        let has_config = git_command()
+            .args(["config", "--get", key])
             .current_dir(dir)
             .output()
-            .map_err(|e| format!("Failed to check git config {key}: {e}"))?;
+            .map(|output| {
+                let value = String::from_utf8_lossy(&output.stdout);
+                output.status.success() && !value.trim().is_empty()
+            })
+            .unwrap_or(false);
 
-        let value = String::from_utf8_lossy(&local.stdout);
-        if local.status.success() && !value.trim().is_empty() {
+        if has_config {
             continue;
         }
 


yikf:
Hi @Giuslock and @lucaronin, could you please review this PR?

This change fixes the issue where Tolaria was always overriding the local Git config with `user.name=Tolaria` and `user.email=vault@tolaria.md`, even when the user had global/system Git config set. The fix now checks for existing config at any scope before setting the fallback values.

Thanks!

Giuslock:
Thanks, this fixes the case where the user already has a global Git identity configured.

One remaining concern: existing vault repositories that already have the local config set to `Tolaria <vault@tolaria.md>` will keep using it, because `git config --get user.email` will find the local value and skip updating it.

Also, for users without any Git config, the fallback is still `vault@tolaria.md`, which may still be associated by GitHub with a real account.

yikf:
@Giuslock Thanks, that makes sense to me. When leveraging Git functionalities, we need to ensure Git is properly configured locally. 

For new users, load configuration values by searching in the order: repo → global → system. 

Existing users follow the same lookup sequence, except we skip any user-level config entry set to vault@tolaria.md and keep searching the next higher scope.

Therefore the final logic is to look up user configurations in the order: repo → global → system. If the value retrieved at the current level equals vault@tolaria.md, skip it and continue searching upwards until a valid configuration is found or an error is raised.

Do you think this is a good idea?

Giuslock:
Yes, that makes sense to me.

Just as a user perspective: the confusing part for me was that `vault@tolaria.md` seems to be associated with another real GitHub account. So when tolaria created commits with that email, GitHub showed those commits as authored by that person inside my private repository.

I understand this is probably only GitHub email-based attribution, not actual access to the repo, but it looked alarming at first.

Maybe it could be worth avoiding `vault@tolaria.md` as a fallback email, and using something non-attributable instead.

Just a suggestion from a user's point of view.

yikf:
cc @LucaRonin , Would you like to share your thoughts?

alepee:
Hi @yikf, thanks for the initial fix and the design discussion here. I went ahead and implemented the full lookup/healing flow you and @Giuslock converged on in #811: it keeps your resolved-scope check, adds healing of vaults that already carry the legacy local `Tolaria <vault@tolaria.md>` pair, skips the legacy email at any scope, and replaces the fallback with a non-attributable `vault@tolaria.invalid` (RFC 2606), plus regression tests for the whole matrix.

The PR description includes a side-by-side comparison with this one. No intent to step on your work: happy to see parts folded into this PR instead, whatever you and the maintainers prefer.

alepee:
Status update for everyone who worked on this thread: the identity fix has landed on `main` via 1309cf32 ("fix: respect configured Git author identity", 2026-06-05). Both this PR (#800) and #811 are now closed.

Lineage, for the record: #800 (@yikf) introduced the core fix, respecting a resolvable git identity instead of always writing `vault@tolaria.md`. @Giuslock surfaced two gaps (healing vaults already carrying the legacy local pair, and a non-attributable fallback). #811 built on that with the heal / respect / fallback rework plus the regression tests. The merged commit does not reference #797, #800 or #811, so I am leaving this pointer here so the discussion stays discoverable from the shipped code.

One technical note that did not make it into the merged version: the fallback email is `vault@tolaria.default`, and `.default` is not a reserved TLD, so it could in principle be registered and a fallback commit attributed to a real account. `.invalid` (RFC 2606 / RFC 6761) is reserved and can never resolve, which is the guarantee the fallback actually needs. Might be worth a one-line follow-up.

yikf:
Appreciate your summary and efforts, @alepee. Well done!

PR #806, opened 2026-06-03
fix: stop serializer inserting trailing-backslash hard breaks and stripping blank lines on save

### LINKED ISSUE #798: Editor serializer inserts trailing backslash hard-breaks and strips blank lines on save
## Summary

When editing a note in Tolaria, the WYSIWYG editor serializes single newlines inside paragraphs as CommonMark hard-break syntax — a trailing `\` followed by a newline — and collapses blank lines I typed. This rewrites my source markdown on every save and is disruptive to a workflow where notes are also edited outside Tolaria or pasted in from external sources.

## Steps to Reproduce

1. Create or open a note in Tolaria.
2. Paste or type a block of prose that contains single newlines between sentences and at least one fully blank line between paragraphs.
3. Save the note (or let autosave fire) and close it.
4. Reopen the file on disk (or in Tolaria).

## Expected

The markdown on disk matches what I wrote: plain newlines stay plain, blank lines remain blank.

## Actual

- Every single newline inside a paragraph is rewritten as `\` + newline (CommonMark hard line break).
- Blank lines I intentionally inserted are removed (or replaced with a lone `\`).
- The effect compounds: any subsequent external edit that re-enters Tolaria gets re-serialized again.

## Example

The note `claude-loop-community.md` in my vault. I pasted clean prose with normal paragraph breaks; after Tolaria saved it, lines 14–50 all end in `\`, and the blank line I added between two paragraphs was stripped.

## Environment

- Tolaria app: `/Applications/Tolaria.app` on macOS (Darwin 24.6.0).
- No git hooks, no sync tool, no external pipeline involved — verified the backslashes appear immediately after Tolaria's own save.

## Request

Please consider one of:

1. A setting to disable hard-break serialization (treat single newlines as soft breaks, i.e. no trailing `\`).
2. Round-trip fidelity: if the source markdown had no trailing `\`, don't add one on save.
3. Preserve blank lines that the user typed between paragraphs.

Happy to test a build or provide more sample files if useful.

mvanhorn:
## Summary

Stops the rich-editor markdown serializer from rewriting users' source on every autosave by stripping BlockNote's lone trailing-backslash soft-break artifact in `compactMarkdown`.

`blocksToMarkdownLossy` emits a CommonMark hard break (a single trailing `\`) for every paragraph-internal line break, even ordinary soft breaks the user typed. `compactMarkdown` is already the post-processor for BlockNote output, so the cleanup lives there: a lone trailing `\` is removed (restoring a plain soft break) while an authored double-backslash `\\` hard break is preserved untouched. Fenced code blocks (existing `inCodeBlock` guard) and indented code blocks (new guard) are left unchanged.

## Why this matters

Issue #798: when a note is edited, the serializer inserts trailing `\` hard breaks on single newlines inside a paragraph. For users who also edit notes outside Tolaria or paste prose from external sources, every autosave mutates the file on disk even when no semantic change was made, churning their source markdown.

## Testing

- `pnpm exec vitest run src/utils/compact-markdown.test.ts` — 35 passed (5 new: soft-break stripping, inline-closer soft breaks, authored double-backslash preservation, indented-code-block preservation, idempotency).
- `pnpm exec eslint src/utils/compact-markdown.ts src/utils/compact-markdown.test.ts` — clean.
- `pnpm exec tsc --noEmit` — clean.
- Full `src/utils/` suite green except one pre-existing, unrelated failure in `releaseDownloadPage.test.ts` (Windows authenticode workflow YAML), which also fails on the unmodified base.

New behavior covered:
- Single newline inside a paragraph serializes without a trailing `\`.
- Authored `\\` hard break is preserved.
- Indented and fenced code blocks are never mutated.
- Idempotent: serializing already-serialized output is byte-identical (no churn on a no-op save).

Fixes #798

AI was used for assistance.


mvanhorn:
The Frontend Tests & Quality Checks failure reproduces on main's own CI (latest main run fails the same workflow), so it isn't introduced by this PR's serializer change.

LucaRonin:
we should have fixed this independently! thank you for your work anyway 🙏

PR #811, opened 2026-06-04
fix(git): never override an existing git identity; heal legacy local config

### LINKED ISSUE #797: Git commits default to vault@tolaria.md, causing GitHub misattribution to another user

alepee:
Fixes #797.

This implements the design discussed in #800 (credit to @yikf for the initial fix and @Giuslock for surfacing the two remaining gaps): respect any resolvable git identity, heal vaults that already carry the legacy hardcoded pair, and make the last-resort fallback non-attributable.

### Problem

`ensure_author_config` only checked the repo-local config before writing `Tolaria <vault@tolaria.md>`. Two consequences:

1. Users with a global gitconfig identity had it silently shadowed: every app-managed commit (autogit, conflict resolution, remote connection, getting started) was authored as Tolaria instead of them.
2. `vault@tolaria.md` is attributable on GitHub (email-based attribution), so those commits showed up as authored by a real third-party account inside users' private repositories.

### Solution

`ensure_author_config` now works in three steps:

1. **Heal**: if the exact legacy `Tolaria <vault@tolaria.md>` pair is still present in the repo-local config (written by earlier versions), remove it so the user's own identity resolves again. A local identity the user set themselves (any other value) is never touched. This migrates existing vaults, addressing @Giuslock's first concern.
2. **Respect**: any identity git resolves from local, global or system scope is kept. A resolved email equal to the legacy fallback is skipped wherever it comes from, per the lookup order discussed in #800.
3. **Fallback**: only when nothing resolves, write a repo-local `Tolaria <vault@tolaria.invalid>`. The RFC 2606 reserved `.invalid` TLD can never be registered on GitHub or any forge, so fallback commits are never misattributed, addressing @Giuslock's second concern.

### How this compares to #800

| Aspect | #800 | This PR |
|---|---|---|
| Respect a global/system git identity (the core fix, from #800) | yes | yes |
| Heal existing vaults that already carry the legacy local pair | no | yes |
| Non-attributable fallback email | no | yes |
| Skip the legacy email regardless of the scope it resolves from | no | yes |
| Propagate git errors when checking config (instead of swallowing spawn failures) | no | yes |
| Regression tests for the heal/respect/fallback matrix | no | yes |
| Identity tests hermetic with respect to the developer's own gitconfig | no | yes |
| Architecture doc updated | no | yes |

### Tests

- New unit tests cover the heal/respect/fallback matrix: global identity respected, legacy pair healed (with and without a global identity), user-set local identity untouched, legacy email skipped even when resolved from the global scope.
- Identity tests now pin `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` to test-controlled files (serialized via a mutex), so they are hermetic with respect to the developer's own gitconfig. Without this, four existing tests fail on any machine that has a global git identity configured.
- Full suite: 973 passed, 0 failed (macOS, stable 1.96.0). `cargo clippy` and `cargo fmt --check` clean.

Side observation, out of scope here: a few non-identity tests spawn raw `git commit` without checking the exit status; on machines with global `commit.gpgsign=true` they can flake under parallel load. Worth a follow-up if maintainers are interested.

### Possible follow-up (UX)

When no identity is resolvable and the user enables git (or autogit), the app could prompt for name/email and write them to the vault's local git config; the `.invalid` fallback would then only cover skipped prompts and headless paths. Happy to work on that as a separate PR if there's interest.


Giuslock:
Fantastic!

yikf:
Thanks a lot for picking up this work and improving upon my original PR https://github.com/refactoringhq/tolaria/pull/800! 

I’ve reviewed your changes, and the complete lookup & healing flow implementation looks really solid and meaningful. 
Appreciate your detailed supplement including legacy vault healing, fallback address adjustment and full matrix regression tests. 

Great job! Thanks @alepee and @Giuslock .

yikf (approved):

alepee:
Closing this PR: the same fix already landed on `main` via 1309cf32 ("fix: respect configured Git author identity"). It uses the same heal / respect / fallback approach for `ensure_author_config`. It heals the legacy repo-local `Tolaria <vault@tolaria.md>` pair, respects any identity resolvable from local/global/system scope (skipping the legacy email), and only writes a repo-local fallback when nothing resolves, with the full heal/respect/fallback test matrix.

One difference worth flagging for a possible follow-up: the merged version uses `vault@tolaria.default` as the fallback email, whereas this PR used `vault@tolaria.invalid`. `.default` is **not** a reserved TLD, so it could in principle be registered and a fallback commit attributed to a real account. `.invalid` is reserved by RFC 2606 / RFC 6761 and can never be registered or resolved, which is exactly the non-attributability guarantee the fallback needs. Switching the fallback to `.invalid` would be a small, safe improvement.

Issue #836, opened 2026-06-08
UI problem on windows


pandawyt:
As shown in the pics, the style's level is wrong when moving the mouse.

<img width="708" height="425" alt="Image" src="https://github.com/user-attachments/assets/f0bb00ef-2c9b-4ea1-bc8e-064219aeb2bf" />

<img width="785" height="415" alt="Image" src="https://github.com/user-attachments/assets/36554792-a2cf-45a3-86f1-08136cbef6da" />

LucaRonin:
Should be fixed in today's stable release! Let me know if you still see it — closing for now.

Issue #827, opened 2026-06-07
Main menu (File, Edit, View) on Windows is unusable


bobrock-git:
Main menu on Windows 11, Tolaria  2026.6.6 need 2 clicks on the each menu label to  unfold.

bobrock-git:
Just checked on another computer.
Ver. 2026.5.29 have the same bug on Windows. Top menu bar doesn't react on first mouse click. It reacts on the second one.

LucaRonin:
Should be fixed in today's stable release! Let me know if you still see it — closing for now.

Issue #826, opened 2026-06-07
Fix Pi CLI discovery and spawning on Windows


kawalem:
### Problem
Tolaria cannot find or execute the Pi CLI (`pi-node`) on Windows.

This is caused by two distinct issues:
1. **Discovery:** The command `where pi` returns both `pi` (a POSIX shell script) and `pi.cmd` (the Windows CMD shim). Tolaria currently picks the first result (the extensionless shell script), which `std::process::Command` cannot execute directly on Windows.
2. **Spawning:** Even if `pi.cmd` is found, passing arbitrary prompt text as command-line arguments to a `.cmd` file fails with the error `"batch file arguments are invalid"`. This happens because `cmd.exe` attempts to interpret special characters within the arguments.

### Root Cause
- In `pi_discovery.rs`, `first_existing_path()` returns the first path from the `where` output without preferring Windows-executable extensions (`.cmd`/`.exe`). (Note: `gemini_discovery.rs` and `opencode_discovery.rs` already handle this correctly).
- In `pi_config.rs`, `build_command()` invokes the binary directly via `hidden_command(binary)`. It should instead use `command_target_avoiding_windows_cmd_shim()` to bypass the `.cmd` shim and call the underlying JS/EXE directly (similar to `gemini_config.rs` and `codex_cli.rs`).

### Suggested Fix
- **`pi_discovery.rs`**: Align `first_existing_path` with the pattern used in Gemini/OpenCode discovery to prefer paths with Windows CLI extensions on Windows platforms.
- **`pi_config.rs`**: Update `build_command` to use `command_target_avoiding_windows_cmd_shim()`. This resolves `pi.cmd` to `node cli.js` and bypasses the `cmd.exe` argument parser entirely.

LucaRonin:
Should be fixed in today's stable release! Let me know if you still see it — closing for now.

PR #842, opened 2026-06-09
fix(kiro): discover and spawn Kiro CLI from Windows npm shims (#826)

evanclan:
## Summary
Extends the Windows npm-shim fix landed for Pi in #826 to Kiro CLI. Kiro was the only agent still using the shared `find_cli_binary` PATH lookup and direct `.cmd` spawning, so Windows installs could be discovered as an extensionless wrapper and then fail to launch with "batch file arguments are invalid".

## Problem
On Windows, `where kiro-cli` returns both an extensionless npm wrapper and `kiro-cli.cmd`. The shared `cli_agent_runtime::first_existing_path` returned the first existing line, and `kiro_cli` spawned that path directly:

```rust
// src-tauri/src/kiro_cli.rs (before)
let mut command = crate::hidden_command(binary);
```

Pi, Codex, Claude, Gemini, and OpenCode already prefer `.cmd`/`.exe` during discovery and resolve shims through `command_target_avoiding_windows_cmd_shim` before passing prompt/chat args.

## Solution
- Teach `cli_agent_runtime::first_existing_path` to skip extensionless wrappers on Windows (same rule as `pi_discovery.rs`).
- Expand `kiro_discovery` candidate paths with Windows npm/pnpm/volta/scoop `.cmd`/`.exe` installs.
- Build Kiro chat commands through `command_target_avoiding_windows_cmd_shim`, matching `codex_cli.rs` / `pi_config.rs`.

## Out of scope
- Pi CLI itself (#826 is already fixed on `main`).
- Other open Windows UI issues (#827, #836).
- Hermes CLI integration (still in draft PR #702).

## Test plan
- [x] `cargo test kiro` — shim resolution + discovery candidate coverage
- [x] `cargo test first_existing` — shared Windows PATH preference regression
- [ ] CI on this PR

## Related
- #826 — Pi CLI Windows discovery/spawn report that documents the same root cause
- #716 — earlier Codex/Claude Windows shim spawn fix

Made with [Cursor](https://cursor.com)


LucaRonin:
Thank you, Evan — this is now implemented directly on `main` in 4a8172a3, with you credited as co-author. Current main already had Kiro chat construction routed through the shared safe command-target resolver, so I kept that implementation and added regression coverage for legacy npm `.cmd` launch. I landed the remaining Windows PATH filtering and package-manager candidates, plus the current official Kiro MSI location under `%ProgramFiles%\\Kiro-Cli`, which Kiro’s newer Windows installer now uses. Validation passed 9 Kiro tests, the full 1,095-test Rust suite at 88.54% line coverage, 5,088 frontend tests, and 27 core smoke tests. Since the production change has landed through Tolaria’s direct-to-main workflow, I am closing this superseded PR.

PR #844, opened 2026-06-09
Enhance frontmatter highlighting with error decoration

krataratha:
Added frontmatterError decoration for invalid YAML syntax highlighting.


LucaRonin:
Thank you, @krataratha — this is now implemented on `main` in cd13657a with your co-authorship credited. I reworked the validation around CodeMirror’s existing Lezer YAML parser so valid nested collections, flow values, comments, and block scalars are not flagged by a line-shape heuristic, and added edit-time regression coverage plus the shared Tolaria error styling. The full push gate passed (5,091 frontend tests, 27 smoke tests, CodeScene 10.0/9.99538). Closing this PR because the improved implementation is already shipped directly to `main`.

PR #887, opened 2026-06-12
fix: route Volta cmd shims through `volta run` (#868)

### LINKED ISSUE #868: Codex CLI chat fails on Windows when PATH resolves to Volta codex.cmd shim
## Summary

On Windows, Codex CLI chat can fail in Tolaria with `Failed to spawn codex: batch file arguments are invalid` when `codex` resolves to a Volta-managed `.cmd` shim such as `C:\Users\Admin\AppData\Local\Volta\bin\codex.cmd`.

The shim usually contains `volta run %~n0 %*`. Tolaria treats that `.cmd` path as the Codex binary and spawns it directly with complex chat arguments.

## Root Cause

Codex chat passes arguments like `exec --json`, sandbox flags, vault paths, MCP `-c` config overrides, paths with spaces such as `C:\Program Files\...`, and the user prompt.

Passing those arguments directly through a Windows batch shim can fail before Codex starts, producing the batch argument error.

Claude may still work on the same machine if Tolaria resolves `claude` to a native executable first, such as `C:\Users\Admin\.local\bin\claude.exe`. In that case Claude avoids the `.cmd` batch shim, while Codex may hit Volta's `codex.cmd`.

## Expected Behavior

Tolaria should launch Codex successfully even when `where codex` resolves to a Volta `.cmd` shim.

## Proposed Fix

Detect Volta-style Windows shims containing `volta run %~n0 %*` and avoid spawning the `.cmd` file directly. Route the command as `volta run codex ...`.

Alternatively, prefer a native `codex.exe` over `.cmd` or `.bat` shims when multiple Codex paths are available.

## Affected Area

- `src-tauri/src/codex_cli.rs`
- `src-tauri/src/cli_agent_runtime.rs`
- `src-tauri/src/cli_agent_runtime/windows_cmd_shim.rs`

## Regression Coverage

Add a Windows-focused regression test where Codex resolves to `codex.cmd`, the shim contains `volta run %~n0 %*`, Tolaria does not spawn `codex.cmd` directly, and the built command is equivalent to `volta run codex ...`.

cat0825:
## Summary

On Windows, Codex CLI chat can fail in Tolaria with `Failed to spawn codex: batch file arguments are invalid` when `codex` resolves to a Volta-managed `.cmd` shim such as `C:\Users\…\AppData\Local\Volta\bin\codex.cmd`.

Such a shim re-dispatches via `volta run "%~n0" %*`. Tolaria's `windows_cmd_shim` only understood npm-style shims (a quoted `%~dp0\…\tool.exe %*` line), so for a Volta shim `target_path()` returned `None` and the runtime fell back to spawning the `.cmd` **directly** with the full chat argument set (`exec --json`, sandbox flags, vault paths, MCP `-c` overrides, paths with spaces, the user prompt). Passing those through a Windows batch shim fails before the agent starts.

Claude may still work on the same machine because it often resolves to a native `claude.exe` first, avoiding the `.cmd` path — which matches the report in #868.

## Fix

- Detect Volta-style shims in `src-tauri/src/cli_agent_runtime/windows_cmd_shim.rs` and route the command as `volta run <tool> …` instead of executing the `.cmd` directly. The `<tool>` name comes from the shim's file stem (e.g. `codex`). When the shim references a resolvable `volta.exe`, that absolute path is used as the program; otherwise it falls back to `volta` on `PATH`.
- Generalize `AgentCommandTarget.first_arg: Option<PathBuf>` into `prefix_args: Vec<OsString>` so a target can carry the two leading arguments Volta needs (`run <tool>`). All call sites updated: codex, claude, pi, opencode, gemini, and the `--version` probe.

This keeps the existing npm/native-exe shim behavior unchanged (verified by the pre-existing test) while adding Volta support.

## Tests

Added regression tests in `windows_cmd_shim.rs`:
- `command_target_routes_volta_codex_shim_through_volta_run` — a `codex.cmd` containing `volta run "%~n0" %*` resolves to program `volta` with prefix args `["run", "codex"]`.
- `command_target_prefers_resolved_volta_exe_when_shim_references_it` — when the shim points at a real `volta.exe`, that absolute path is used.

Verified locally (macOS host):
- `cargo test --lib` for the touched modules — all pass (`windows_cmd_shim`, `codex_cli`, `claude_cli`, `pi_config`, `opencode_config`, `gemini_config`, `cli_agent_runtime`).
- `cargo clippy --lib` — clean.

The Windows-specific spawn path can only be exercised end-to-end on Windows; the logic under test (shim parsing → command target) is platform-independent and covered by the unit tests above. Leaving the native pre-push lane (Playwright smoke / CodeScene / coverage gates) to CI.

Fixes #868


LucaRonin:

PR #906, opened 2026-06-15
fix: prevent Korean IME list item splitting

dhr05074:
## Summary

Fixes an editor IME edge case where Korean composition at the start of a list item can trigger an unwanted paragraph/list split.

## Root cause

The existing IME guard stopped composing `Enter` keydown events before BlockNote/Tiptap list shortcuts handled them, but WebKit can still emit a post-composition `beforeinput` event with `inputType: insertParagraph`. That path was not guarded, so the list item could split even though the composing keydown had been intercepted.

## Changes

- Track the last `compositionend` timestamp inside the IME composition key guard.
- Capture `beforeinput` and suppress one immediate `insertParagraph`/`insertLineBreak` while the editor is composing or just after composition ends.
- Keep normal Enter/list editing behavior available outside the composition window.
- Add unit and smoke coverage for the post-composition paragraph insertion path.

## Validation

- `pnpm exec vitest run src/components/imeCompositionKeyGuardExtension.test.ts` - passed, 13 tests
- `pnpm exec playwright test tests/smoke/korean-list-ime-composition.spec.ts --reporter=line` - passed, 1 test
- `pnpm lint` - passed
- `npx tsc --noEmit` - passed
- `pnpm build` - passed
- `git diff --check` - passed
- `git status --short -- demo-vault demo-vault-v2` - clean
- `pnpm l10n:validate` - passed, 20 locale catalogs against 816 English keys
- `pnpm test:coverage --silent` - passed, 410 files / 4501 tests, line coverage 88.15%

## Release notes

- Localization: no UI copy changes.
- PostHog: no event needed because this is low-level editor input correctness.
- Docs/ADR: none, no architecture or public API change.
- CodeScene/Codacy: local CodeScene/Codacy MCP/CLI credentials were unavailable in this session; CodeScene remains enforced by PR CI via `.codescene-thresholds`.


LucaRonin:
Thank you, Juchan — this fix is now implemented directly on `main` in 8aa0bcb8, with you credited as co-author. I kept the post-composition `beforeinput` guard but tightened it into a one-shot causal sequence: only a composing Enter arms suppression, the matching paragraph insertion consumes it, and a deliberate normal keydown cancels it. That preserves normal Enter immediately after composition. I also repaired the stale Playwright fixture signature, then verified 17 focused unit tests, both Korean IME browser scenarios, the full 5,088-test coverage gate, and 27 core smoke tests. Since the production change has landed through Tolaria’s direct-to-main workflow, I am closing this superseded PR.

PR #927, opened 2026-06-20
feat: add update_note and append_to_note MCP tools

### LINKED ISSUE #895: MCP: Add `update_note` and/or `append_to_note` tools
## Summary

The current MCP server exposes `create_note` for writing to the vault but has no way to update an existing note. This makes agent-assisted workflows effectively append-only — agents can create new notes but cannot maintain living documents that evolve over time.

## Current tool surface

The 8 current tools are: `search_notes`, `get_vault_context`, `list_vaults`, `get_note`, `create_note`, `open_note`, `highlight_editor`, `refresh_vault`.

Notable gap: `get_note` can read a note's full content, but there's no corresponding write-back.

## Use cases this blocks

- **Maintaining agent/project notes** — an agent note (e.g. an `Agent` type note tracking an AI assistant's configuration) needs updating when its status, focus, or schedule changes. Today that requires the user to edit manually in Tolaria even when an agent has all the information to do it.
- **Evolving project notes** — a `Project` note accumulates decisions, milestones, and status over time. Without `update_note`, agents can only create supplementary notes, which fragments the knowledge graph.
- **Keeping note metadata current** — frontmatter properties like `status`, `schedule`, or `related_to` wikilinks change as projects evolve. There's no MCP path to update them.

### `update_note`
Replace the full content (frontmatter + body) of an existing note.

```json
{
  "name": "update_note",
  "description": "Replace the content of an existing vault note. The path must already exist — use create_note for new notes.",
  "inputSchema": {
    "type": "object",
    "properties": {
      "path": {
        "type": "string",
        "description": "Vault-relative path to the existing note (e.g. NanoClaw/ionga.md)"
      },
      "content": {
        "type": "string",
        "description": "Full new file content (frontmatter + body)"
      }
    },
    "required": ["path", "content"]
  }
}
```

### `append_to_note` (optional but useful)
Add content to the end of an existing note without replacing the full file. Safer for agents since it avoids accidentally overwriting content.

```json
{
  "name": "append_to_note",
  "inputSchema": {
    "type": "object",
    "properties": {
      "path": { "type": "string" },
      "content": { "type": "string", "description": "Markdown to append after existing body" }
    },
    "required": ["path", "content"]
  }
}
```

## Notes

- If conflict protection is a concern, an optional `expectedMtime` or etag field on `update_note` would allow agents to fail-safe rather than blindly overwrite concurrent edits.
- `append_to_note` is lower-risk than `update_note` and covers a large share of the use cases (changelogs, log sections, follow-up bullets). Could ship first as a stepping stone.
- The vault is iCloud-synced — both tools should call `refresh_vault` (or equivalent) after writing so the in-memory index stays current.

nguyenngothuong:
Closes #895. The Tolaria MCP server previously only let agents create new notes (`create_note` with `O_EXCL`). This PR adds write-back to existing notes so agents can maintain living documents (Agent notes, evolving Project notes, frontmatter `status` / `schedule` / `related_to` updates) instead of fragmenting them into supplementary notes.

## What was implemented

Two new writable, non-destructive MCP tools operating on **existing** notes, layered exactly like `create_note` (vault.js primitive → tool-service method → stdio + WebSocket registration):

- **`update_note`** — replace full content (frontmatter + body). Atomic temp-file + `rename` so a crashed write never truncates a note. Optional `expectedMtime` (the `mtimeMs` from `get_note`) fails fast on concurrent edits, leaving the on-disk note untouched.
- **`append_to_note`** — append markdown verbatim to the body. No conflict guard because append cannot lose data on its own. Lower-risk path for logs/journals.

Both reuse the existing `realpath` traversal guard, resolve via the same `writableVaultPath` / `writableNotePath` helpers, emit `vault_changed` so the UI refreshes, and use the existing writable annotation profile (`readOnlyHint: false`, `destructiveHint: false`, `openWorldHint: false`) — i.e. MCP clients treat them like `create_note` (non-destructive local write, no extra approval).

Design follows the spec in #895 exactly. The optional `expectedMtime` conflict guard is opt-in (issue suggestion), so the common single-editor agent flow stays a single `update_note` call without a mandatory read.

## QA

Verified via MCP stdio integration tests:
- `update_note` / `append_to_note` are listed with correct annotations and required `['path', 'content']` schema.
- Round-trip `update_note` then read-back matches the new content.
- `append_to_note` appends verbatim (no auto-newline — agents control formatting).
- `update_note` rejects path-traversal outside the vault; fails on `expectedMtime` mismatch without touching the note.
- `append_to_note` rejects path-traversal; throws `ENOENT` for missing notes.

No native UI QA needed — no UI change (these are MCP-only tools).

## Tests / coverage

`cd mcp-server && node --test test.js tool-service.test.js` → **53 pass / 0 fail** (8 new cases across vault primitives, service layer, and stdio MCP integration). `pnpm lint` clean. Pre-push gate: 25/25 Playwright smoke pass.

## CodeScene

PAT / project not set in this environment, so the gate skipped. Boy Scout rule applied manually — `vault.js` helpers (`writeExistingUtf8File`, `appendUtf8File`, `assertNoteMatchesExpectedMtime`) are focused single-purpose functions, and the atomic write cleans up its temp file on a failed rename so it never leaks.

## Codacy

No Critical / High findings introduced — all writes go through the existing `resolveVaultNotePath` traversal guard (no new path handling); `expectedMtime` is opt-in and never silently overwrites.

## Localization

No UI copy changes.

## PostHog

No event needed — MCP tool calls are agent-driven, not user-driven UI actions; instrumenting them would create noise.

## Refactoring

Extracted `resolveWritableNote` and `noteContent` helpers in `tool-service.js` to keep the two new methods parallel with `createNote` without duplication.

## ADRs

ADR-0142 records the design and the alternatives considered (mandatory `expectedMtime`, frontmatter-patch tool, in-place write, append-only shipping).

## Docs

`mcp-server/index.js` tool banner updated to list the new tools. Bundle regenerated (`pnpm bundle-mcp`) — bundle is gitignored, rebuilt at release time.

## Demo vault dirt

`git status --short -- demo-vault demo-vault-v2` clean.


LucaRonin:
Thank you for the PR! We merged this and it should now be available in the latest stable release. Closing this out.

PR #958, opened 2026-06-25
fix: render bare attachments/ image paths (#956)

### LINKED ISSUE #956: Support image paths without ./ prefix (Obsidian compatibility)
## Problem

Standard markdown image syntax `alt` without a `./` prefix does not render in the rich editor. Images only render when the path includes a `./` prefix:

- ❌ `alt` — does not render
- ✅ `alt` — renders correctly

Both are valid standard markdown relative paths. The difference matters because tools like Obsidian generate paths without the `./` prefix by default.

## Steps to Reproduce

1. Create a note with an image in the `attachments/` folder
2. Write `test` in the note
3. Image does not render in the rich editor
4. Change to `test`
5. Image renders correctly

## Expected Behavior

Tolaria should render standard markdown image syntax `alt` without requiring a `./` prefix. Both forms are equivalent in standard markdown and should work identically.

## Impact

This affects interoperability with other markdown editors (Obsidian, Logseq, etc.) that generate `attachments/...` paths without the `./` prefix. Users migrating from these tools or working across multiple editors will see broken images.

## Suggested Fix

In the markdown preprocessing step before BlockNote parsing, normalize bare `attachments/` paths by adding a `./` prefix. Something like:

```typescript
markdown.replace(
  /(!\[[^\]]*\]\()((?!\.\/|\.\.\/|https?:\/\/|asset:)[a-zA-Z])/g,
  "$1./$2"
)
```

Or alternatively, modify `isPortableAttachmentPath` / `resolvePortableAttachmentUrl` in `vaultImages.ts` to also handle paths starting with `attachments/` — which it already does in the URL resolution layer. The issue is that BlockNote parser does not create image blocks from these paths without the `./` prefix.

kshanxs:
Resolves #956. Normalizes bare relative image paths to include the `./` prefix in the editor markdown preprocessor, and updates path resolution to handle both forms consistently. Also includes safety checks to preserve UNC/backslash Windows paths and filesystem absolute paths.


kshanxs:
@copilot please fix the merge conflicts in this pull request.

LucaRonin:

PR #975, opened 2026-06-29
fix: apply user-select: none to non-editor UI chrome

### LINKED ISSUE #905: UI needs generous helping of `user-select: none`
Adding user-select none is a low effort high impact polish to make the app feel like real app and not a web page.  I notice it randomly when clicking buttons or fields and it just selects the text in the button.

Image

mvanhorn:
## Summary
Add a base rule to the global stylesheet so the app root defaults to `user-select: none` (with the `-webkit-user-select` prefix for the Tauri/WebKit webview), then re-enable `user-select: text` for the surfaces where selection must work: the editor/content area, text inputs, textareas, `[contenteditable]`, and any code/preview panes. Place the global default and the editor/input opt-back-in rules in `src/index.css` (which already owns base and theme contracts) and adjust any conflicting chrome layout rules in `src/App.css` if they need the explicit opt-out. Keep the change presentation-only: no component/TSX logic changes, so the editing and copy-paste experience in the note body is preserved while button/label text no longer highlights on click.

## Why this matters
Tolaria's UI chrome (buttons, fields, sidebar items, toolbars) lets the OS text-selection highlight trigger on click, so clicking a button or label often selects its text and the app feels like a web page instead of a native desktop app. The reporter asks for a generous application of `user-select: none` across non-editor UI, calling it a low-effort, high-impact polish. Maintainer LucaRonin (COLLABORATOR) replied on 2026-06-16 "you are 100% right - will do", endorsing the change; no PR has materialized in the 13 days since. `user-select` currently appears in only a handful of files (notably `src/components/EditorTheme.css`) and there is no app-wide chrome reset.

See https://github.com/refactoringhq/tolaria/issues/905.

## Testing
- Happy path: clicking a sidebar item, toolbar button, or property label no longer selects its text. - Editor preserved: selecting and copying text inside the note body / markdown editor still works. - Inputs preserved: selecting text inside search boxes, the AI chat input, and frontmatter fields still works. - Selectable content preserved: AI panel message history and code blocks remain selectable for copy.

Fixes #905

AI was used for assistance.


LucaRonin:

PR #976, opened 2026-06-29
fix: parameterize settings shortcut in footer for cross-platform display

Urielbp:
## Summary

- The settings footer shortcut hint hardcoded `⌘,` directly in every locale string, which is macOS-only and displays incorrectly on Windows (`Ctrl+,`) and Linux
- Replaced the hardcoded symbol with a `{shortcut}` placeholder in all 18 locale files
- `SettingsFooter.tsx` now resolves the shortcut at render time via the existing `formatShortcutDisplay` utility, so each platform sees its native shortcut representation

## Changes

- `src/components/SettingsFooter.tsx` — import `formatShortcutDisplay`, compute `SETTINGS_SHORTCUT` once at module level, pass it as the `shortcut` interpolation param to `t('settings.footerShortcut', ...)`
- `src/lib/locales/*.json` (18 files) — replace hardcoded `⌘,` with `{shortcut}` placeholder

### Before (Linux)
full page before
linux before

### After (Linux)
full page after
linux after

## Testing

- ESLint pre-commit hook passed on all changed files
- TypeScript compilation verified (no `tsc` errors introduced)
- No logic changes — purely a string parameterization using the existing i18n interpolation system already used elsewhere in the app

> AI was used for assistance in authoring this contribution.


Urielbp:
Just tested on macOS and it's working as intended.
Captura de Tela 2026-06-29 às 12 23 51

PR #984, opened 2026-07-02
fix(breadcrumb): close tooltip when pointer leaves icon (#982)

### LINKED ISSUE #982: [Bug] The breadcrumb bar above the editor hovering-display bug.
> (p.s.) This is just a little display bug, not very serious.
1. When I move my cursor to the icons in the breadcrumb bar above the editor, it automatically shows the details below as expected. 
2. However, when I move the cursor between two icons, the explanation may do not change to the right content. 
3. It often happens when I slide directly and slowly from one to the other one.
4. The bug may occurs among several buttons and not among others.
Apology for my poor English :)

JuizSpeaking:
Closes #982.

### Bug

When the cursor slid slowly from one toolbar icon to an adjacent one, the previous tooltip would stick on screen instead of switching to the new icon's tooltip.

### Root cause

`useBreadcrumbTooltipControl` activated the tooltip label on both `onPointerEnter` and `onPointerMove`. During a slow slide from icon A to icon B, the browser can deliver a trailing `pointermove` on A *after* `pointerenter` on B has already fired, which re-set the active label back to A and left A's tooltip on screen.

The existing test for adjacent-icon tooltip switching was a simplified sequence that only fired `pointerMove(B)` and never exercised the trailing move on the previous icon — so the race wasn't covered.

### Fix

Drop `onPointerMove` from the activation path. The active label is now driven by `pointerenter` / `pointerleave` only. A real browser slide always fires `pointerleave` on the outgoing icon and `pointerenter` on the incoming one, so the active label transfers cleanly and a trailing `pointermove` on the outgoing icon can no longer override it.

`focus` / `blur` still activate / deactivate for keyboard users, so the keyboard path is unchanged.

### Tests

- New regression test for the slow-continuous-slide sequence.
- New regression test for a trailing `pointermove` on the previous icon after the cursor has already entered the next icon (the exact #982 repro).
- Updated existing adjacent-icon test to use `pointerleave` + `pointerenter` (what a real browser fires), instead of the simplified `pointerMove`.

### Verification

- 4840/4840 frontend tests pass.
- `pnpm lint` clean.
- `tsc --noEmit` clean.
- `vite build` clean.
- Frontend coverage: 88.46% lines / 75.92% branches (≥70% gate).
- Rust checks skipped (no `src-tauri/` changes).
- Playwright smoke + CodeScene health gate deferred to CI (not runnable in the temp clone used to author this PR).


LucaRonin:

PR #1001, opened 2026-07-06
fix: frontmatter title extraction fails on Windows CRLF, archive filter hardcoded

fix2015:
Two bugs:

- \`extract_fm_title_value\` in rename.rs only checked for \`---\\n\` (Unix line endings) but not \`---\\r\\n\` (Windows CRLF). On Windows, the rename system couldn't extract frontmatter titles, falling back to filename-based extraction and producing incorrect wikilink updates. The main frontmatter parser in \`ops.rs\` already handles both formats — matched the rename code to the same pattern.

- \`evaluateView\` hardcoded \`!e.archived\` before evaluating filter conditions, making it impossible to create a view that shows archived entries. Even if a user sets up a filter like \`archived equals true\`, it always returns zero results because the hardcode eliminates them first. Removed the hardcode so the filter condition system handles archived status like any other field.


LucaRonin:
Thank you for the PR! We merged your contribution, and it is now available in the latest stable Tolaria release. Thanks again for helping improve Tolaria.

PR #1081, opened 2026-07-29
fix: stop sheet notes being overwritten with the previous note's body (#1080)

### LINKED ISSUE #1080: Data loss: `_display: sheet` notes are overwritten with the active editor note's body
## Summary

A note with `_display: sheet` gets its body replaced by the body of whatever note is currently active in the editor. The sheet's own frontmatter is preserved, only the body is replaced. The CSV content is destroyed with no warning and no undo.

The two notes do not need to be related in any way. In the minimal reproduction below there is no wikilink, no expression, and no reference of any kind between them.

I lost a spreadsheet's contents four times over about 30 minutes. It was only recoverable because I started committing to git after the second occurrence.

## Environment

- Tolaria 2026.7.22 (CFBundleVersion 2026.7.22)
- macOS 26.5.2 (25F84), Apple Silicon
- Vault is a local git repository, no remote
- Reproduced with a freshly deleted cache (see "Ruled out" below)

## Reproduction

**1.** Create a plain note, `probe-a-self.md`:

````md
---
type: Note
probe_value: hello
---
# Probe A self

```html height="80"
<p>A: {{probe_value}}</p>
```
````

**2.** Create a sheet note, `probe-sheet.md`:

```md
---
type: Note
_display: sheet
---
Name,Wert
x,42
```

**3.** Open the vault in Tolaria.

**4.** Open `probe-a-self`. It opens in the normal editor.

**5.** Switch to `probe-sheet`. It opens in the spreadsheet editor.

**6.** Click **"Return to editor"**.

`probe-sheet.md` on disk is overwritten immediately.

## Expected

`probe-sheet.md` is untouched. It has nothing to do with `probe-a-self`.

## Actual

`probe-sheet.md` on disk becomes:

````md
---
type: Note
_display: sheet
---
# Probe A self

```html height="320"
<p>A: {{probe_value}}</p>
```
````

The frontmatter is `probe-sheet`'s own, preserved. The body is `probe-a-self`'s, copied in. The CSV body (`Name,Wert` / `x,42`) is gone. `probe-a-self.md` itself is unchanged.

Both notes then appear in the sidebar under the same title, "Probe A self", since both files now carry the same H1.

## Notes on the trigger

The trigger is step 6, "Return to editor", and both steps 4 and 5 are required. Opening the sheet without having viewed another note first does not reproduce it.

A guess at the mechanism, from the outside: "Return to editor" switches a `_display: sheet` note from the spreadsheet editor into the rich text editor, and the rich text editor still holds the document from the previously viewed note. That document is then saved to the sheet note's path.

If that is right, the body written to the sheet should always be the body of the last note viewed in the normal editor, which matches every occurrence I saw.

In every occurrence the victim was a `_display: sheet` note and the copied body came from the previously active note. Plain notes were never affected: in the reproduction above, three other plain notes in the same vault were untouched.

One earlier occurrence also involved renaming a sheet note from within Tolaria. That rename left an empty `.tolaria-rename-txn/` directory in the vault root, which may be a separate issue.

## Secondary effect: render error

Once a sheet's body is no longer CSV, any HTML block referencing a cell in it throws and takes down the editor pane:

```
Tolaria render error

L1@tauri://localhost/assets/Editor-DK5LTvKx.js:529:1675
DW@tauri://localhost/assets/Editor-DK5LTvKx.js:1716:7872
Zqe@tauri://localhost/assets/Editor-DK5LTvKx.js:1716:27434
map@[native code]
Uqe@tauri://localhost/assets/Editor-DK5LTvKx.js:1716:28394
@tauri://localhost/assets/Editor-DK5LTvKx.js:1716:28754
map@[native code]
Gqe@tauri://localhost/assets/Editor-DK5LTvKx.js:1716:28661
Ph@tauri://localhost/assets/index-Bp3w3zXU.js:23:57007
OUe@tauri://localhost/assets/Editor-DK5LTvKx.js:1719:2266
OUe@tauri://localhost/assets/Editor-DK5LTvKx.js:1719:2174
render@tauri://localhost/assets/Editor-DK5LTvKx.js:1727:1451
HtmlBlock@tauri://localhost/assets/Editor-DK5LTvKx.js:320:148079
```

Minimal case: an HTML block containing `<p>{{[[some-sheet]].B12}}</p>` where `some-sheet` has been clobbered and no longer has a `B12`.

This may be worth fixing independently. `reference/vault-expressions.md` states that unresolved expressions "remain visible as escaped placeholders ... so broken dashboards fail visibly instead of silently showing the wrong value". A thrown exception that blanks the editor is not that, and it made the underlying data loss much harder to diagnose, since the crash looked like the primary bug.

## Ruled out

Each of these was tried and the overwrite still recurred:

- Deleting `~/Library/Caches/club.refactoring.tolaria` and `~/Library/WebKit/club.refactoring.tolaria` (twice)
- `Reload Vault` from the command palette
- Fully quitting the app between restores
- Removing the duplicate H1 from the sheet (the shared title is a result of the copy, not its cause)
- Renaming the sheet file from the shell so it shared no lineage with the other note
- Removing all wikilinks and expressions connecting the two notes

## Workaround

Avoid "Return to editor" on `_display: sheet` notes. Every occurrence I could reproduce went through that action.

## Impact

Silent, repeatable destruction of spreadsheet contents. There is no in-app undo once the file is rewritten, and a user without version control loses the data outright. The failure is easy to miss because the frontmatter survives, so the note still looks like a sheet in listings.

### PR DESCRIPTION
**Pull request by sim-san**
Fixes #1080.

## The bug

A `_display: sheet` note's body was silently replaced by the body of the previously viewed note. Frontmatter survived, the CSV did not, and there was no undo. The reporter lost a spreadsheet four times in about thirty minutes.

## Root cause

The rich BlockNote editor is a **single shared instance**, and `useRichEditorSheetSwapState` deliberately never loads sheet notes into it — `richEditorActiveTabPath = null`, `tabsForEditorSwap = []` (`src/components/useRichEditorSheetTransition.ts:43`). While a sheet is active, the editor therefore still holds the *previously viewed note's* document.

Entering raw mode ran `syncActiveTabIntoRawBuffer`, which serialized that stale document and combined it with the **active tab's** frontmatter:

```ts
const [frontmatter] = splitFrontmatter(tabContent)   // the sheet's frontmatter
return `${frontmatter}${bodyMarkdown}`               // the previous note's body
```

(`src/utils/richEditorMarkdown.ts:168`)

The intended guard — `serializeRichEditorContent`, false when there is no pending editor change — was reopened by:

```ts
const shouldSerializeRichEditorContent =
  serializeRichEditorContent || hasRichEditorDurableBlocks(editor.document)
```

That second term is true whenever the *stale* document contains an html, mermaid, tldraw, callout, or file-attachment block. This is why the reported repro needs an html block, and it explains the detail that pinned the diagnosis down: the reporter's authored `height="80"` came back as `height="320"`, which is `HTML_BLOCK_DEFAULT_HEIGHT` (`src/utils/htmlBlockMarkdown.ts:12`). The body was re-serialized from BlockNote blocks, not copied from the file.

## The fix

Reuse the ownership guard the codebase already applies elsewhere — `activeEditorChangePath` (`useEditorTabSwap.ts:207`) and `cachePreviousTabOnPathChange` both check `editorContentPathRef.current !== path` before touching the editor. The raw-mode flush simply didn't.

- `editorContentPathRef` is hoisted into `Editor.tsx` and shared between `useEditorTabSwap` and `useRawModeWithFlush`
- `syncActiveTabIntoRawBuffer` now returns the tab's own content untouched when the rich editor does not hold that note

`useEditorTabSwap` still creates its own ref when none is passed, so no call sites outside `Editor.tsx` change. This closes the general hazard, not only the sheet case.

## One correction to the issue report

On `main`, clicking **Return to the editor** *alone* does not write — the `content === rawInitialContent` guard at `editorRawModeSync.ts:154` holds. The corrupted buffer is created on **entering** raw mode and persists to disk on the next flush: ⌘S, a note action, or a tab switch (`App.tsx:515`, `App.tsx:1097`). Root cause and severity are unchanged; only the exact moment of the write differs from the report.

## Testing

New regression test `src/components/Editor.sheetRawRoundTrip.test.tsx`, verified **red without the guard and green with it**. It drives the real seam: note A active → switch to sheet → toggle raw → flush → assert the sheet is never written with note A's body.

- ESLint `--max-warnings=0`: pass
- `tsc --noEmit`: pass
- Vitest: 500 files / 5203 tests pass
- Frontend coverage: 85.44% statements, 88.91% lines (gate ≥70%)
- Codacy: 0 findings on all five touched files, before and after; `codacy:gate` passes
- CodeScene: **not run** — no MCP server, no `cs` CLI, and `CODESCENE_PAT` / `CODESCENE_PROJECT_ID` unset in the authoring environment. Needs verification before merge.

## Out of scope

Two other findings in #1080 are untouched here:

1. Unresolved cell expressions (`{{[[some-sheet]].B12}}`) throw and blank the editor pane with "Tolaria render error", contradicting `reference/vault-expressions.md`. This materially slowed the diagnosis.
2. An in-app sheet rename leaves an empty `.tolaria-rename-txn/` directory behind.

## Architectural note

The shared BlockNote instance is ambient state whose "which note do I hold?" invariant is re-implemented independently at four call sites. Three check it; this one forgot. A single accessor that refuses to serialize unless ownership matches would make this class of bug unrepresentable.


LucaRonin:
Thank you @sim-san. Your PR identified the root cause and directly informed the implementation shipped in the [stable Tolaria v2026-08-11 release](https://github.com/refactoringhq/tolaria/releases/tag/v2026-08-11). We adapted the ownership guard to the narrower current-main synchronization seam, added regression coverage and related transaction cleanup, and shipped it in `8b5f24e0f76cd594c0b30a45865c38ad27ecccfe`. We appreciate and credit your original contribution here. Closing the PR now that the task is complete.

PR #1100, opened 2026-08-03
feat(icons): expose the full Phosphor set (287 → 1,530) instead of a curated list

jspectre-ui:
### Problem

`src/utils/iconRegistry.ts` hand-maintains 288 named imports. Any icon outside that list can't be used as a type icon, and adding one means editing the registry. `@phosphor-icons/react` already ships the rest.

### Solution

Enumerate the library's exports and derive kebab-case names, instead of curating them.

One wrinkle worth flagging: `@phosphor-icons/react` exports **every icon twice** — `Acorn` and `AcornIcon`. A naive enumeration registers both, filling the picker with ~1,500 duplicates. This keeps the bare name and drops the `*Icon` alias where its twin exists.

**287 → 1,530 selectable icons**, no duplicates, no new dependency, −280 lines.

### Compatibility

Public API is unchanged — `ICON_OPTIONS`, `findIcon`, `resolveIcon`, `IconEntry`, `IconProps` keep their signatures. `TypeCustomizePopover`, `noteIcon`, and `typeIcon` need no changes. Non-icon exports (contexts, providers, base components) are filtered out; names are de-duplicated and sorted.

### Testing

- `pnpm vitest run` — **498 files, 5,195 tests, all passing**
- `pnpm tsc --noEmit` — clean
- `src/utils/iconRegistry.test.ts` passes unmodified (`gear-six`, `cooking-pot`, null/unknown → `FileText`)
- Pre-commit hook (`eslint . --max-warnings=0 --quiet`) passes

### Tradeoff

The bundle grows because every icon is reachable. That seemed right for a desktop app, but if you'd rather keep it lean this could be gated behind a lazy or dynamic import — happy to rework it that way.

I've been running this in a personal build for about a month; the duplicate-alias issue above is something I only caught while preparing this PR, so the fix is included here rather than in what I'd been using.


LucaRonin:
Thank you @jspectre-ui. Your PR directly informed the full Phosphor icon-catalog implementation shipped in the [stable Tolaria v2026-08-11 release](https://github.com/refactoringhq/tolaria/releases/tag/v2026-08-11). We adapted and improved the design with lazy icon loading and progressive picker rendering to reduce bundle size, shipping it in `b29d59d497f95ee1bb2d99529d9ad7f9117f9dec`. We appreciate and credit your original contribution here. Closing the PR now that the task is complete.

PR #1031, opened 2026-07-13
fix(editor): prevent Chinese IME duplication in nested ordered lists

hedgeho9X:
Closes #1029

## Summary
- guard Space as an IME composition commit key alongside Enter
- prevent BlockNote list shortcuts from splitting or duplicating a composing list item
- add unit and Playwright regression coverage for ordered lists nested under check lists
- document the expanded IME composition guard

## Validation
- pre-push lint and production build passed
- 5,018 frontend tests passed
- coverage passed (88.71% lines)
- 27 core Playwright smoke tests passed
- targeted Chinese IME composition Playwright regression passed

## Local analysis availability
- CodeScene was skipped by the repository hook because its credentials are not configured locally
- Codacy CLI/integration was not available locally

Issue #1029, opened 2026-07-13
[Bug] Chinese IME input is duplicated in ordered lists nested under task lists


hedgeho9X:
**Description:**

In Tolaria, when an ordered list is nested under a task list (`- [ ]`), entering Chinese text with a Chinese input method and pressing Space to confirm the input causes both the pinyin and the confirmed Chinese text to be inserted.

**Steps to Reproduce:**

1. Create a task list item:

   ```text
   - [ ] 123
   ```

2. Create an ordered list under the task list item.
3. Use a Chinese input method to type `nihao`.
4. Press Space to confirm the input.
5. Repeat the process with additional Chinese text.

**Actual Result:**

The pinyin and the confirmed Chinese text are both inserted. For example:

```text
- [ ] 123
  1. ni'hao
  2. 你好
```

**Expected Result:**

Only the confirmed Chinese text should remain after pressing Space:

```text
- [ ] 123
  1. 你好
```

**Environment:**

- Tolaria version: `2026.07.09` — latest stable version
- Operating system: macOS `26.5`
- Chinese input method: WeChat Input Method
- Is this issue limited to ordered lists nested under task lists? Yes

Screenshots are attached for reference.

<img width="900" height="666" alt="Image" src="https://github.com/user-attachments/assets/00f25531-8028-4171-a2ce-b9235f66d0e9" />

LucaRonin:
Should be fixed in today's stable release! Let me know if you still see it — closing for now.

PR #1182, opened 2026-08-28
fix(editor): preserve Safari IME commits in nested lists

ZackYJz:
## Summary

- keep Safari nested-list text blocks alive during IME composition with an out-of-model ProseMirror widget decoration
- wait for ProseMirror’s post-composition change before persistence, with a fallback for IMEs that emit no final change
- add regression coverage proving the sentinel is never serialized and persistence observes the reconciled commit

## Root cause

Safari/WebKit commits composition by first emitting `deleteCompositionText`. When the composing text is the only text node in an empty nested list item, ProseMirror can normalize away that node and move the selection before `insertFromComposition` arrives. The committed CJK text then lands in a new list item while the preedit pinyin is restored in the original item. A leading space masks the bug because the text block never becomes empty.

This is the same WebKit/ProseMirror behavior tracked in [ProseMirror #934](https://github.com/ProseMirror/prosemirror/issues/934), and is a follow-up to #1029 / #1031.

## Validation

- `pnpm vitest run src/components/imeCompositionKeyGuardExtension.test.ts src/components/SingleEditorView.test.tsx` — 59 tests passed
- ESLint passed for all four changed files
- `pnpm build` passed
- full `pnpm test` passed — 526 test files / 5,339 tests, plus 5 release-version tests
- native packaged macOS app tested with Sogou Pinyin:
  - nested ordered list: `ceshi` committed as `测试` in the same item, with no retained pinyin or extra row
  - nested bullet list: same result
  - saved Markdown matched the rendered editor state

No dependencies, UI copy, analytics events, or public APIs are changed.


LucaRonin:
Thank you, @ZackYiJun — I validated the Safari IME approach against current `main` and landed an adapted version directly in `e501eac6977faedc1e8f3232030a26ba17a921c9` (`fix: preserve Safari IME nested-list commits`). The implementation retains your authorship via `Co-authored-by: Zack YiJun <zackyjz@qq.com>` and adds focused regression coverage for sentinel lifecycle plus the 20 ms reconcile / 50 ms fallback behavior. Full frontend, Rust coverage, focused IME Playwright, core smoke, CodeScene, and Codacy gates passed. Since the equivalent fix is now on `main`, I’m closing this PR as superseded.

PR #1195, opened 2026-08-31
fix: repair mcp-server bundle crash on launch (missing policy file + empty import.meta)

XavierWangHX:
## Problem

The shipped mcp-server (`Contents/Resources/mcp-server/index.js`) crashes immediately on every launch, so MCP registration fails in every client (Claude Code, Cursor, Gemini CLI, pi, ...) with `Connection closed`. Running the bundle directly shows:

```
TypeError: Invalid URL
    at new URL (node:internal/url:899:25)
    at Object.<anonymous> (.../Resources/mcp-server/index.js:28531:36)
  code: 'ERR_INVALID_URL',
  input: './app-config-policy.json'
```

Two independent bugs in `scripts/bundle-mcp-server.mjs`:

1. **`app-config-policy.json` never ships.** `vault-path.js` reads it via `readFileSync(new URL('./app-config-policy.json', ...))` at module load, and the file exists in `mcp-server/` in the repo, but the bundling script only writes `package.json`, `index.js` and `ws-bridge.js` into `src-tauri/resources/mcp-server/`, so it never lands in the `.app` resources.
2. **CJS output breaks `import.meta.url`.** With `format: 'cjs'`, esbuild rewrites `import.meta` to an empty `var import_meta = {};` stub (it also warns about this), so `new URL('./app-config-policy.json', import_meta.url)` receives `undefined` as base and always throws `ERR_INVALID_URL` — regardless of Node version, cwd, or client.

## Fix

- Copy `app-config-policy.json` from `mcp-server/` into the bundle output next to `index.js` / `ws-bridge.js`.
- After each esbuild build, rewrite the `var import_meta = {};` stub to resolve from the bundle's real location at runtime: `require("node:url").pathToFileURL(__filename).href`. If esbuild ever changes the stub format, the script fails loudly instead of shipping a broken bundle.

(esbuild's `define` can't express this — define values must be an entity name or JS literal, not a call expression.)

## Verification

- `pnpm bundle-mcp` → output contains the policy file, both bundles carry the patched `import_meta`, zero stubs remain.
- `pnpm tauri build` → built `Tolaria.app`; an MCP `initialize` + `tools/list` handshake against `Tolaria.app/Contents/Resources/mcp-server/index.js` succeeds (server v0.3.0, 12 tools registered).
- Before the fix the same handshake fails with the crash above.
- `eslint scripts/bundle-mcp-server.mjs` clean; no new biome findings (the three pre-existing `useNodejsImportProtocol` hints in this file are untouched).


LucaRonin:
Thank you for the PR! We couldn't use this implementation in the final solution, but the task it addressed is now complete and available in the latest stable Tolaria release. Thanks again for taking the time to contribute.

Issue #1236, opened 2026-09-17
Sheet Cell Content Disappears


Reagan1947:
## 1.1 Issue Description
I copied the contents from an Excel spreadsheet into a Tolaria spreadsheet. **When I click to edit the spreadsheet and move the focus**, the **contents of the cells disappear**.

## 1.2 Steps to Reproduce the Issue
1. Open the Excel spreadsheet  [test_sample_sheet.xlsx](https://github.com/user-attachments/files/32324477/test_sample_sheet.xlsx)  in the attachment, select the main content in the spreadsheet, and copy it into the spreadsheet file you created in Tolaria.
2. Double-click to select any cell that contains data. Enter Edit Mode
3. Double-click other cells to edit them

**Expected Results**
The content that was previously in the cell is still there

**Current Situation**
The previous cell contents were cleared

## 1.3 Test Versions and Systems
Tolaria: Alpha 2026.9.12.1
Computer Systems: Macos 27 (26A428)

## 1.4 Screenshots as Evidence

**Edit Cell:**

<img width="721" height="447" alt="Image" src="https://github.com/user-attachments/assets/2f3e3543-0d42-4cca-983f-9b3fe5bb44f8" />

**Content Disapper:**

<img width="780" height="395" alt="Image" src="https://github.com/user-attachments/assets/5b3baa1d-f23e-442c-a395-9535df744968" />

## 1.5 Related Issue
#1155: Sheet notes on Windows: title becomes the full file path, and cell edits are lost or delete existing content

#1080: Data loss: _display: sheet notes are overwritten with the active editor note's body

LucaRonin:
Fixed by #1237 (squash-merged to main). Thanks @Reagan1947 — keeping this open until the fix ships in a stable, per release cleanup.

LucaRonin:
Should be fixed in today's stable release! Let me know if you still see it — closing for now.

PR #1237, opened 2026-09-17
fix: preserve sheet cell content on blur

Reagan1947:
Fix sheet cell content disappearing after focus change
---
## Summary
- Prevent committed sheet cell content from being overwritten with an empty value when focus moves to another cell.
- Validate that the editor value still matches the last committed value before handling a subsequent blur event.
- Add a regression test covering the pointer-down, editor reset, selection change, and blur sequence.
## Root Cause
After IronCalc committed a cell edit, it cleared its internal editor value. The following blur event treated that programmatic reset as a new edit and wrote an empty string back to the original cell.
Testing
- Added a regression test for switching focus after editing an existing cell.
- Verified the behavior manually in the sheet editor.
- 5,404 frontend tests passed.
- Frontend line coverage: 89.53%.
- Rust tests and the 85% coverage threshold passed.
- ESLint, TypeScript checks, and the production build passed.

## Related Issue
Sheet Cell Content Disappears https://github.com/refactoringhq/tolaria/issues/1236


LucaRonin:
Thank you for the PR! We merged your contribution, and it is now available in the latest stable Tolaria release. Thanks again for helping improve Tolaria.

PR #1242, opened 2026-09-18
fix: preserve settings scroll position

aicayzer:
## Summary

Saving an immediate settings toggle previously remounted the inner settings panel. That replaced its scroll container and sent the user back to the top of the page.

This change keeps the panel mounted while still synchronizing the editable draft with saved settings. As a result, navigation within Settings preserves the existing scroll position after an immediate toggle updates the saved configuration.

## Tests

- Added a regression test that toggles “Show images”, then verifies the scroll container is unchanged and retains its scroll position.
- Pre-commit checks passed.

## Notes

Version 2026.9.8

## Evidence

https://github.com/user-attachments/assets/8ec2f39b-fb6e-4811-88ba-65c51548409b


LucaRonin:
Thanks for the contribution! This looks like something Tolaria already supports today — `SettingsPanel` on `main` no longer remounts the inner panel with a `key={JSON.stringify(initialDraft)}`, so toggling an immediate setting keeps the same scroll container.

I’m going to close this for now to keep the queue tidy, but please reopen or comment if I misunderstood the gap.

LucaRonin:
Closing because settings scroll is already fixed on `main`.

aicayzer:
Ahh thanks – bug shown in screenshot seems to be fixed.

